PR #31, opened 2020-02-18
add clusterrolebinding for controller service account

patrickshan:
add missing clusterrolebinding for controller service account

### DIFF
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -10,3 +10,18 @@ rules:
   - apiGroups: ["projectcontour.io"]
     resources: ["httpproxies"]
     verbs: ["get", "list", "create", "update", "delete", "deletecollection", "patch", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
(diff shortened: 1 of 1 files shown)


config/200-clusterrole.yaml
```diff
@@ -10,3 +10,18 @@ rules:
   - apiGroups: ["projectcontour.io"]
     resources: ["httpproxies"]
     verbs: ["get", "list", "create", "update", "delete", "deletecollection", "patch", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
mattmoor: We actually shouldn't need this 🤔 

In Serving we use an [aggregated cluster role](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) [here](https://github.com/knative/serving/blob/89d45259bbc65a29f142d3c7d6179a86cc865477/config/core/200-serviceaccount.yaml#L23-L33) which should pick up the role above simply by having it labelled.

patrickshan: didn't know aggregated cluster role. That's very neat. This PR can be closed.

PR #34, opened 2020-02-19
Take config-network HTTPProtocol in consideration when listing targets

skaslev:
Address a TODO in the probe target lister.


pkg/reconciler/contour/lister.go
```diff
@@ -43,7 +43,20 @@ var _ status.ProbeTargetLister = (*lister)(nil)
...
+
+	var port int32
+	var scheme string
+	switch cfg.Network.HTTPProtocol {
+	case network.HTTPEnabled:
+		port, scheme = 80, "http"
+	case network.HTTPRedirected:
+		port, scheme = 443, "https"
+	case network.HTTPDisabled:
+		return results, nil
```
mattmoor: This should be same as Redirected I think.

skaslev: Done.

pkg/reconciler/contour/lister.go
```diff
@@ -43,7 +43,15 @@ var _ status.ProbeTargetLister = (*lister)(nil)
 func (l *lister) ListProbeTargets(ctx context.Context, ing *v1alpha1.Ingress) ([]status.ProbeTarget, error) {
 	var results []status.ProbeTarget
 
-	visibilityKeys := config.FromContext(ctx).Contour.VisibilityKeys
+	cfg := config.FromContext(ctx)
+
+	port, scheme := int32(80), "http"
+	switch cfg.Network.HTTPProtocol {
+	case network.HTTPDisabled, network.HTTPRedirected:
```
shashwathi: Can you add test case for this configuration?

pkg/reconciler/contour/lister.go
```diff
@@ -63,35 +71,33 @@ func (l *lister) ListProbeTargets(ctx context.Context, ing *v1alpha1.Ingress) ([
...
 			urls = append(urls, &url.URL{
-				Scheme: "http",
+				Scheme: scheme,
 				Host:   host,
 			})
 		}
 
-		// TODO(mattmoor): Perhaps key off of whether HTTP is enabled?
-		portName, err := network.NameForPortNumber(service, 80)
+		portName, err := network.NameForPortNumber(service, port)
```
patrickshan: need to make sure `port` is int32 here. so I think it's better to cast it to int32 here instead of on line #48

pkg/reconciler/contour/lister.go
skaslev: Done.

pkg/reconciler/contour/lister.go
skaslev: `port` gets defined as int32 on #48. The untyped `443` constant on #51 is implicitly converted to int32 on assignment[1].

[1] https://golang.org/ref/spec#Representability

PR #35, opened 2020-02-20
Update Contour and adjust default classes to avoid collisions

mattmoor:
The `config/contour` is now wholly populated by `./hack/update-deps.sh`

Fixes: https://github.com/knative/net-contour/issues/10


hack/update-deps.sh
```diff
@@ -28,3 +28,89 @@ dep ensure
...
+  sed -e $'s@hostPort:@# hostPort:@g'
+}
+
+function privatize_loadbalancer() {
+  sed "s@type: LoadBalancer@type: ClusterIP@g" \
+    | sed "s@externalTrafficPolicy: Local@# externalTrafficPolicy: Local@g"
+}
+
+
+# TODO(mattmoor): DO NOT SUBMIT
```
mattmoor: Found "DO NOT SUBMIT".

PR #90, opened 2020-04-02
Remove duplicate cluster role binding definitions

andrew-su:
When creating contour-internal and contour-external, the cluster role
binding from the contour project is left intact. This generates two
copies of the `ClusterRoleBinding`s with the same name pointing to
different subjects. However, this is not needed as we define our own
cluster role bindings for internal and external.

related issue: https://github.com/knative/net-contour/issues/53

/cc @shashwathi

### DIFF
--- a/config/contour/external.yaml
+++ b/config/contour/external.yaml
@@ -1390,19 +1390,6 @@ spec:
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
-kind: ClusterRoleBinding
-metadata:
-  name: contour
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: contour
-subjects:
-- kind: ServiceAccount
-  name: contour
-  namespace: contour-external
----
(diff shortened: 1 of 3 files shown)


andrew-su:
The `sed` command doesn't seem to work on mac but it does on linux (tested on ubuntu). If anyone has a better suggestion, please recommend.

andrew-su:
`sed -e '/apiVersion: rbac/{' -e ':a' -e 'N' -e '/---/!ba' -e '}' -e '/kind: ClusterRoleBinding/d'` will be able to run on Mac however, on top of deleting the ClusterRoleBinding, it also deletes the `RoleBinding`:
```
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: RoleBinding
metadata:
  name: contour-leaderelection
  namespace: projectcontour
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: contour-leaderelection
subjects:
- kind: ServiceAccount
  name: contour
  namespace: projectcontour
```

On Linux, it works as expected.

I tried running these commands on the `02-rbac.yaml` file from contour example.

PR #116, opened 2020-05-11
Convert from godep to go modules

andrew-su:
We need to ensure that the generated yaml is correct for release.

/cc @shashwathi


mattmoor:
/lgtm
/approve

Awesome! 🎉

PR #118, opened 2020-05-16
Change go.mod to use 1.13, so ingress tests actually runs

tcnghia:
Currently there is a bug in test-infra such that when no integration tests is run we won't see a failure.

Using go 1.14 in go.mod hits this condition.  See https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_net-contour/117/pull-knative-net-contour-integration-tests/1261318878126936064  .  

Change back to 1.13 so tests actually will run.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module knative.dev/net-contour
 
-go 1.14
+go 1.13
 
 require (
 	github.com/google/go-cmp v0.4.0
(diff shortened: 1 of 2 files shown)


tcnghia:
@andrew-su it seems like some dependencies need to be pinned  ?

```
# knative.dev/net-contour/pkg/client/clientset/versioned/scheme
pkg/client/clientset/versioned/scheme/register.go:34:2: undefined: "github.com/projectcontour/contour/apis/projectcontour/v1".AddToScheme

andrew-su:
That's an issue which is fixed by applying the patch. Run `hack/update-deps.sh`

markusthoemmes:
@tcnghia do you mind rebasing this? Contour nightlies are broken because of it! :)

PR #121, opened 2020-05-19
Align test calls with net-istio.

markusthoemmes:
This should unblock the integration test failures for now.


markusthoemmes:
The issue here now is twofold:
1. I need to not use the vendor dir to be able to get the upstream tests.
2. I need to use the vendor dir to be able to observe the patch we applied.

Ugh.

PR #123, opened 2020-05-19
[RELEASE] Pin knative.dev/{pkg,test-infra} to release-0.15

markusthoemmes:


shashwathi:
@markusthoemmes  : Looks like ObservedGeneration is updated when there is a failure in reconciler. We need to update unit tests as this changes behaviour. 

change ingress status to include: `i.Status.ObservedGeneration = 1`  in [here](https://github.com/knative/net-contour/blob/master/pkg/reconciler/contour/contour_test.go#L235)

PR #124, opened 2020-05-19
Add `networking.knative.dev/ingress-provider: contour` label.

tcnghia:
This will allow serving operator to have a way to pick and choose ingress controller for installation.


shashwathi:
@tcnghia : Why do we need labels on rbac and other contour resources?

shashwathi:
I just looked at net-istio repo. Looks like all resources under config gets the label. Got it 👍

PR #135, opened 2020-06-02
Run `yq` dependency from vendor rather than `go get` and `go install`

andrew-su:
We don't need the version to keep updating.

This also works around the issue with failing nightly builds as it was complaining about dependency not being marked explicit.

See slack: https://knative.slack.com/archives/C012J5TCS6Q/p1591062361000800 for more details about the CI failures.

### DIFF
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -64,15 +64,7 @@ rm -rf $(find vendor/ -path '*/e2e/*_test.go')
 # Add permission for shell scripts
 chmod +x $(find vendor -type f -name '*.sh')
 
-function install_yq() {
-  if [ -x "$(command -v yq)" ]; then
-    return
-  fi
-  go get github.com/mikefarah/yq/v3
-}
-
-# install yq
-install_yq
+go install ${ROOT_DIR}/vendor/github.com/mikefarah/yq/v3
 
 function add_ingress_provider_labels() {
   sed '${/---/d;}' | yq m - ./hack/labels.yaml -d "*"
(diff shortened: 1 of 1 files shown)


hack/update-deps.sh
```diff
@@ -59,6 +59,16 @@ mkdir -p test/conformance/ingress
 rm -rf test/conformance/ingress/*
 rsync -r --no-perms "$SERVING_MODULE"/test/conformance/ingress/ test/conformance/ingress
 
+# Install yq command if it doesn't exist
+function install_yq() {
+  if [ -x "$(command -v yq)" ]; then
+    return
+  fi
+  go get github.com/mikefarah/yq/v3@3.3.0
```
mattmoor, line 67: Should this be in hack/tools.go?

shashwathi: We do not need any files from yq dependency so I dont think this should be part of hack/tools.go. WDYT?

mattmoor: The point of `hack/tools.go` is to pull binary dependencies used in scripts like this into vendor (e.g. the K8s code-generators).  IMO we should follow a similar pattern of letting `go.mod` be the source of truth for version / pinning and changing this to just build the binary for use below.

shashwathi:
@mattmoor : ready for another review

hack/update-deps.sh
```diff
@@ -64,15 +64,7 @@ rm -rf $(find vendor/ -path '*/e2e/*_test.go')
...
-function install_yq() {
-  if [ -x "$(command -v yq)" ]; then
-    return
-  fi
-  go get github.com/mikefarah/yq/v3
-}
-
-# install yq
-install_yq
+go install ${ROOT_DIR}/vendor/github.com/mikefarah/yq/v3
```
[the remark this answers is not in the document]
andrew-su: Fair point, we'll just run it instead of installing things on people's machines :)

shashwathi:
@mattmoor : Ping

Issue #12, opened 2020-01-29
TestUpdate is pretty flaky


mattmoor:

mattmoor:
See also: https://github.com/projectcontour/contour/issues/2151

mattmoor:
I believe that the linked PR works around the core issue with Contour here, but there are lingering issues in Knative that need to be addressed before this will pass reliably.

Tracing through the prober to understand how it was succeeding without having successfully probed, I noticed that it was actually abandoning probes due to 400s, which after peeling away at things turned out to be the issue fixed here: https://github.com/knative/serving/pull/8764

I suspect that there may be some additional lingering issues around some atomicity assumptions in the probing logic, but that's still theoretical and unconfirmed to be a problem in practice.

mattmoor:
I'm going to disable this test again since it seems to still fail more often than not.

mattmoor:
Here's a log dump from a failure during TestUpdate that shows the new endpoint probe becoming ready and things still failing when the new routing is rolled out after that...

```
# This is where we queue probes for the endpoints.

    TestIngressConformance/4/update: kubelogs.go:200: I 16:21:41.864 contour-ingress-controller-8699d87d89-gkjtx [status/status.go:265] [serving-tests/ingress-conformance-4-update-yphpbidh--ep] Queuing probe for http://ingress-conformance-4-update-iamjjxtu.ingress-conformance-4-update-yphpbidh.serving-tests.net-contour.invalid, IP: 10.52.2.195:80 (depth: 0)
    TestIngressConformance/4/update: kubelogs.go:200: I 16:21:41.864 contour-ingress-controller-8699d87d89-gkjtx [status/status.go:265] [serving-tests/ingress-conformance-4-update-yphpbidh--ep] Queuing probe for http://ingress-conformance-4-update-iamjjxtu.ingress-conformance-4-update-yphpbidh.serving-tests.net-contour.invalid, IP: 10.52.3.39:80 (depth: 0)
    TestIngressConformance/4/update: kubelogs.go:200: I 16:21:41.864 contour-ingress-controller-8699d87d89-gkjtx [status/status.go:265] [serving-tests/ingress-conformance-4-update-yphpbidh--ep] Queuing probe for http://ingress-conformance-4-update-iamjjxtu.ingress-conformance-4-update-yphpbidh.serving-tests.net-contour.invalid, IP: 10.52.4.76:80 (depth: 0)
[...]

Issue #165, opened 2020-07-22
Guard DeleteCollection in Lister lookup


mattmoor:
_Relates to the ideas behind: https://github.com/knative/pkg/issues/1520_

We should not be making unguarded API calls during Reconciliation, otherwise resyncs become a "thundering herd" of API calls.  Here we have an unguarded DeleteCollection call, which *even if it deletes nothing* eats into our rate limit: https://github.com/knative-sandbox/net-contour/blob/434b08b5a1c88735ec2e886214448f3dbc8a2a99/pkg/reconciler/contour/contour.go#L155

We have a lister cache already, it should be relatively trivial to guard the DeleteCollection in a call to check our cache for stuff to delete prior to the call.

/kind good-first-issue

PR #166, opened 2020-07-22
Add a scheme for warming Endpoints before rolling our kingress changes.

mattmoor:
This is based on the sketch here: https://docs.google.com/document/d/11k0d0STquDFYBp_43I2AQQVDqQ22ZWfnon7RmJBcc2s/edit#

Contour has a race when it rolls out programming to Envoys where the routing
data can land prior to the endpoint data.  This is due to the fact that Contour
filters our Endpoints that aren't referenced by the routing layer to reduce the
amount of information it distributes, however, this means that when migrating
traffic to a new service that the Envoys have not seen before, the Envoys may
be directed to send traffic to Endpoints for which it has not yet received data.

This problem is especially bad for Knative because every Revision rolls out a
service that the Envoys haven't seen before, so virtually every rollout risks
a blip of 503s, which we actually see manifest fairly regularly in the e2e
testing, and which the TestUpdate kingress conformance test has exacerbated
for a while.

The solution in this pull request builds on the prior work to probe the Envoys
to ensure programming has been distributed.  It breaks the kingress
reconciliation down into two discrete phases:
1. Distribute artificial programming that references all of the Endpoints of
  our new programming.  Once probing has confirmed that to be Ready, then we...
2. Distribute the actual programming that references the new Endpoints, which
  we have confirmed the Envoys are aware of.

The implementation of this two-phased approach is done via a shallow recursion.
We create a secondary kingress targetting the appropriate Envoy pools with bogus
hostnames, and an annotation to block further recursion.  Once this "endpoint
probe" kingress has reported Ready, then we proceed with the remained of the
Contour reconciliation as we had it before, but now with the guarantee that the
Endpoints data is present.

Fixes: https://github.com/knative-sandbox/net-contour/issues/12

This should fix the longstanding TestUpdate issue, which I ran successfully (in
isolation) a half dozen times last night.

Related to: https://github.com/projectcontour/contour/issues/2151

I've tried to limit the scope of this change, but there are a handful of follow
ups I intend to make once this starts baking:

1. https://github.com/knative-sandbox/net-contour/issues/165: I noticed this
  making this change and it means that we are violating the "no calls on nops"
  rule for reconcilers.

2. Clean up the endpoints probe kingress resources.  I added a bool for this and
  a TODO, but I want to circle back to that once this has landed.  Technically
  what this PR does is sufficient for correctness, but it will scale better if
  we don't leave around the garbage programming.

3. Clean up the legacy Endpoints checking, which is subsumed by this and is now
  unnecessary complexity.

### DIFF
--- a/pkg/reconciler/contour/contour.go
+++ b/pkg/reconciler/contour/contour.go
@@ -21,14 +21,18 @@ import (
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 
 	contourclientset "knative.dev/net-contour/pkg/client/clientset/versioned"
 	contourlisters "knative.dev/net-contour/pkg/client/listers/projectcontour/v1"
+	ingressclientset "knative.dev/networking/pkg/client/clientset/versioned"
 	ingressreconciler "knative.dev/networking/pkg/client/injection/reconciler/networking/v1alpha1/ingress"
+	networkingv1alpha1 "knative.dev/networking/pkg/client/listers/networking/v1alpha1"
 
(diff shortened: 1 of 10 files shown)


mattmoor:
These all timed out, so I'm going to try cleaning my cluster and seeing what I might have missed

mattmoor:
Found the problem.  Resource names were getting too long, and failing validation 😉 

I switched to using `kmeta.ChildName` and things are making progress.  Will push once my confidence it higher.

PR #167, opened 2020-07-22
Clean up the endpoint probe kingress after the parent is ready.

mattmoor:
This wasn't strictly necessary for the prior change, so I split it off.


mattmoor:
Hrm....  I'm still seeing occasional 503s in the continuous runs for update too 🤔

mattmoor:
This is disabled now... 😢 

/retest

PR #177, opened 2020-07-27
[master] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/assign n3wscott shashwathi tanzeeb
/cc n3wscott shashwathi tanzeeb


dprotaso:
I think you have to add yourself to this list

https://github.com/knative/community/blob/master/peribolos/knative-sandbox.yaml

looks like tanzeeb was lost in the transition

Issue #176, opened 2020-07-27
TestRouteCreation still fails with 503s


mattmoor:
Looking through the Envoy debug logs for one of these failures, the 503 correlates with the following:
```
[2020-07-27 00:12:13.603][16][debug][router] [source/common/router/router.cc:464] [C33723][S16000592596892839825] unknown cluster 'serving-tests/route-creation-luxnutmg-8kxhs/80/da39a3ee5e'
```

What's interesting is that this cluster is added to the Envoy pod serving the 503 at `00:12:09` (add) to `00:12:11` (remove).

Then it is added back at `00:12:13:593` (starts 10ms before the 503, but not done until 5ms after the 503).

What I find puzzling here is that the endpoint probe should cover this, and should properly overlap these, but it doesn't seem to be happening 🤔 

I'm not sure we have adequate logging to track this down currently, but it was relatively straightforward to reproduce this with:
```
go test -v -tags=e2e -count=10 -timeout=20m ./test/conformance/api/{v1alpha1,v1beta1,v1} -run=TestRouteCreation --ingressClass=contour.ingress.networking.knative.dev
```

Things to look into when this happens:
1. Is the endpoint probe kingress being deleted sooner than we think?
2. Are the clusters Contour produces for the endpoint probe the same as produced for the actual kingress?
3. Is the test code somehow flip/flopping the kingress spec in some way?

mattmoor:
Key to debugging these things is focusing on the Envoy logs.  They should be set to debug here: https://github.com/knative-sandbox/net-contour/blob/d61a8cee37e4e14a1211456b1c9a3ae5774d394e/config/contour/external.yaml#L1787

The jumping off point for `503`s is to search for a `503 NR` (generally ignore the `/healthz` requests with `K-Network-Probe`, these are our readiness probes).  You can then look at timestamp of the log statement (often skewed from where it's recorded by logging systems, e.g. stackdriver).  Then look for an "unknown cluster" log statement with a similar embedded timestamp (these drift apart because of stdout vs. stderr buffering).

Once you have the cluster name, you can filter on that and see how it comes/goes from the debug logs.

PR #184, opened 2020-08-05
Fix a bad assumption in the HTTP Proxy logic.

mattmoor:
The net-contour logic for translating kingress to httpproxy resources is a one-to-many
translation.  In fact, the arity of the http proxies can *change* dynamically over the
lifetime of a kingress.  So the logic to query, create, update, and delete httpproxy
resources is heavily label based.

Based on how Knative creates kingress resources, I do not believe it is possible to hit
this problem today, but given how we recursively use kingress for endpoint probing,
combined with how Knative uses kingress to invoke us, we hit this interesting corner case.

Essentially, the logic as it exists today assumes that the tuple `(kingress name, domain)`
is enough to uniquely identify the httpproxy; however, if a single kingress exposes the
same domain at multiple visibilities then this is insufficient.  We need the tuple
`(kingress name, domain, visibility)`.

Generally this wasn't a problem before because the external domain `foo.bar.example.com`
went to one visibility, while the other domain `foo.bar.svc.cluster.local` went to the other.
However, for the endpoint probe kingress, we create a host per service so that our probing
confirms that every endpoint is available.  Given that knative exposes the service by both
`foo.bar.example.com` AND `foo.bar.svc.cluster.local` this created a problem where we would
attempt to create the same endpoint probe httpproxy differing only in visibility.

There are two aspects to the fix for this:
1. Incorporate the visibility (class) into the name of the http proxy.
2. Label the http proxies with visibility, and filter on visibility when choosing which
  pre-existing httpproxy to update.

Despite changing resource names, this should be rollout safe because it only affects the
endpoint probe resource, which is new, and that resource doesn't exist in the steady state.

Fixes: https://github.com/knative-sandbox/net-contour/issues/176

### DIFF
--- a/pkg/reconciler/contour/contour.go
+++ b/pkg/reconciler/contour/contour.go
@@ -188,6 +188,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) r
 		selector := labels.Set(map[string]string{
 			resources.ParentKey:     proxy.Labels[resources.ParentKey],
 			resources.DomainHashKey: proxy.Labels[resources.DomainHashKey],
+			resources.ClassKey:      proxy.Labels[resources.ClassKey],
 		}).AsSelector()
 		elts, err := r.contourLister.HTTPProxies(ing.Namespace).List(selector)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


mattmoor:
cc @tcnghia

tcnghia:
/hold 

for https://github.com/knative/serving/pull/8931 confirmation

mattmoor:
/unhold

Much better 😈

PR #185, opened 2020-08-06
Disable endpoint probing with path prefixes.

mattmoor:
Right now net-contour doesn't properly support certmanager-based http01 challenges because it expects all references services to adhere to the dataplane contract around probing.

Related: https://github.com/knative-sandbox/net-certmanager/issues/44

### DIFF
--- a/pkg/reconciler/contour/resources/kingress.go
+++ b/pkg/reconciler/contour/resources/kingress.go
@@ -60,12 +60,19 @@ func MakeEndpointProbeIngress(ctx context.Context, ing *v1alpha1.Ingress, previo
 		}
 
 		for _, route := range proxy.Spec.Routes {
+			hasPath := false
+			for _, cond := range route.Conditions {
+				if cond.Prefix != "" {
+					hasPath = true
+				}
+			}
 			for _, svc := range route.Services {
 				si, ok := sns[svc.Name]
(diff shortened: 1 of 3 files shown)


mattmoor:
I pulled this in here to verify it fixes the auto-TLS tests: https://github.com/knative/serving/pull/8931

pkg/reconciler/contour/resources/kingress.go
```diff
@@ -94,6 +101,10 @@ func MakeEndpointProbeIngress(ctx context.Context, ing *v1alpha1.Ingress, previo
 
 	for _, name := range l {
 		si := sns[name]
+		if si.HasPath {
```
ZhiminXiang, line 104: not very familiar with contour. Does this mean the service that has path won't be added into the route rules? Then will the http01 challenge service still be accessible and be able to serve http01 challenge traffic?

mattmoor: Ha, no.  This is a *recursive* kingress that creates a host for each service that we can probe to ensure that all Envoys know about all of the Endpoints before we roll out new route programming.

We can't do this for HTTP01 challenges (the only thing we use Path for) because certmanager doesn't support proper probe responses.

PR #195, opened 2020-08-11
Have Envoy emit debug logs

mattmoor:


dprotaso:
Do the debug logs print changes to the envoy config?

mattmoor:
> Do the debug logs print changes to the envoy config?

Not exactly, but you can generally infer key changes from this.  The info logs print some information around EDS, but I've found that RDS is pretty muted without this.

This may not be super useful without an improves logging story in Contour itself to help connect the dots, but it's better than nothing.

dprotaso:
worst case - there's an admin endpoint that could be used to dump the config at some interval

https://www.envoyproxy.io/docs/envoy/latest/operations/admin#get--config_dump

mattmoor:
@dprotaso yeah, I debated hitting that, but the problem is that these things are typically super racy and we'd have to know which envoy to hit and when.  If we can find a good way to leverage that, I'm all ears!

PR #224, opened 2020-08-23
Configure default retries matching Istio

mattmoor:
Every so often we see a failure like [this](https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-contour-latest/1296639201952927744), where we get:
```
    util.go:131: Saw unexpected response "upstream connect error or disconnect/reset before headers. reset reason: connection failure" 1 times.
```
_(thanks to @vagababov for the improved instrumentation!)_

It turns out that when we dropped the explicit retries in the kingress spec, we allowed Istio's default retry policy to kick in, which insulates net-istio from this sort of thing, but Contour doesn't have similar defaults.  With this change, we explicitly configure a default similar to Istio's.


mattmoor:
Also updated networking to pull in the change to address 404s.

PR #228, opened 2020-08-27
[master] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/assign n3wscott shashwathi tanzeeb
/cc n3wscott shashwathi tanzeeb


mattmoor:
I think this is broken because when I rolled back to 1.4 I didn't run `--upgrade`, so the clients didn't rollback, and there were some breaking changes I need to fix up.

I'm trying a few things in parallel, including trying to get us onto the most recent version of contour that will pass serving's testing.  When I get that sorted out I'll try to circle back and fix this.

mattmoor:
Alright this is fixed here: https://github.com/knative-sandbox/net-contour/pull/233

Issue #203, opened 2020-08-13
Avoid reprobing Ready kingress


mattmoor:
On each reconciliation we have an unguarded call to statusManager.IsReady here: https://github.com/knative-sandbox/net-contour/blob/c988b46b9ac4d89e869533a9cd9b4e0493c6c5ec/pkg/reconciler/contour/contour.go#L239

Generally this library call is idempotent and once the controller has reached a steady state it should be quite cheap, however, in the case of failovers, all existing keys must be probed to warm the probers cache, which is (in theory) overkill considering they may already be ready.

Consider reducing the cost of failovers (and rollouts, and ...) by eliding the initial probe for already-Ready kingress.

Issue #226, opened 2020-08-25
Latest net-contour is struggling with some serving tests


mattmoor:
See: https://github.com/knative/serving/pull/9125

mattmoor:
I think the main change in this was the switch to Contour 1.7, though a few other changes have accumulated since.

mattmoor:
1.8 was just released: https://github.com/projectcontour/contour/releases/tag/v1.8.0

mattmoor:
Ok, new staged PR to repro the break: https://github.com/knative/serving/pull/9200

mattmoor:
I refactored scale test a bit to try and make debugging things a bit easier.

mattmoor:
After investigating this a bit, it looks like we are never able to successfully probe certain routes.  I haven't gone deeper into why yet.

I saw at least one instance where we failed-over mid probe due to chaos, and it looks like the failover took the full 15s for there to be a new lease holder (likely need the K8s patch for leaderelection).  For good measure, I'm trying scale-100 with chaos disabled, but I don't think it's the cause because after the failover the probes pick up and are still seeing 404s.

In an effort to bring us as far forward as I can, and narrow the search for potentially breaking Contour changes, I have opened PRs to try against contour 1.5 and 1.6 as well (we know 1.7 hit this too, since that's what I tried first).

mattmoor:
1.5: https://github.com/knative/serving/pull/9209
1.6: https://github.com/knative/serving/pull/9210
1.8: https://github.com/knative/serving/pull/9200

mattmoor:
Looks like it started in 1.5 :(

mattmoor:
Started a thread on this in Contour slack: https://kubernetes.slack.com/archives/C8XRH2R4J/p1598722245143600

mattmoor:
Looking at a failure with a bit more instrumentation...

I see the following probe failures immediately prior to things timing out:
```
    kubelogs.go:197: E 18:37:42.981 contour-ingress-controller-577bf7cc7f-r66m5 [status/status.go:397] [serving-tests/scale-to-n-scale-100-069-of-100-vyyiymqy] Probing of http://scale-to-n-scale-100-069-of-100-vyyiymqy.serving-tests failed, IP: 10.16.13.3:80, ready: false, error: unexpected status code: want 200, got 404 (depth: 0)
```

I see this hostname first queued for probing ~90 seconds prior:
```
[...]

PR #231, opened 2020-08-28
Avoid thundering herd of probes during failover.

mattmoor:
This change guards the status prober with a simple readiness check as an
optimization for the case of failing over and globally resyncing resources.
When a fresh controller pod slurps through ~100 kingress for the first time
we will currently probe ALL of them.  On a 10 Node cluster, with 1 public and
3 private hostnames, this works out to a lot of probes:

```
  100 * (1 * 10 + 3 * 10) = 4000
```

Factor in what we do with endpoint probing, and it adds roughly 2000 probes.

Currently, this is what the scale-100 test in serving is doing every 30s or so,
in addition to the standard work it is doing.

Fixes: https://github.com/knative-sandbox/net-contour/issues/203

Possibly related to: https://github.com/knative-sandbox/net-contour/issues/226

### DIFF
--- a/pkg/reconciler/contour/contour.go
+++ b/pkg/reconciler/contour/contour.go
@@ -142,6 +142,8 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) r
 		}
 
 		if !actualChIng.IsReady() {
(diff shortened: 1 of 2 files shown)


mattmoor:
Ok, this clearly has bugs.

Issue #237, opened 2020-09-02
upstream connect error or disconnect/reset before headers


mattmoor:
See: https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-sandbox-net-contour-continuous/1300971558608572416

A 503 is served during "dispatch/percentage":

```
    util.go:886: Error meeting response expectations: got unexpected status: 503, expected map[200:{}]
    util.go:886: HTTP/1.1 503 Service Unavailable
        Transfer-Encoding: chunked
        Content-Type: text/plain
        Date: Wed, 02 Sep 2020 02:13:17 GMT
        Server: envoy
        Vary: Accept-Encoding
        
        5b
        upstream connect error or disconnect/reset before headers. reset reason: connection failure
        0
```

mattmoor:
Another one [here](https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-sandbox-net-contour-continuous/1300624270317064192) during "visibility/split":

```
    util.go:886: Error meeting response expectations: got unexpected status: 503, expected map[200:{}]
    util.go:886: HTTP/1.1 503 Service Unavailable
        Content-Type: text/plain
        Date: Tue, 01 Sep 2020 03:18:29 GMT
        Server: envoy
        Vary: Accept-Encoding
        X-Envoy-Upstream-Service-Time: 269
        
        upstream connect error or disconnect/reset before headers. reset reason: connection failure
```

mattmoor:
I'm wondering is this will help: https://github.com/knative-sandbox/net-contour/pull/232/commits/25362a16e040d8aac61cce4d6c73189326487b3d

mattmoor:
Doesn't look like it: https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-sandbox-net-contour-continuous/1301469716732186624

```
    util.go:886: Error meeting response expectations: got unexpected status: 503, expected map[200:{}]
    util.go:886: HTTP/1.1 503 Service Unavailable
        Transfer-Encoding: chunked
        Content-Type: text/plain
        Date: Thu, 03 Sep 2020 11:02:03 GMT
        Server: envoy
        Vary: Accept-Encoding
        
        54
        upstream connect error or disconnect/reset before headers. reset reason: local reset
        0
```

mattmoor:
Interestingly the message seems to have shifted from `reset reason: connection failure` to `reset reason: local reset`.

Here are three more from recent kingress conformance runs:
[...]

PR #246, opened 2020-09-08
Adjust the default RetryOn policy.

mattmoor:
This adjusts the default retry policy to more correctly match Istio's, and add one new kind of retry that Istio does not seem to do: `reset`.

WRT "more correctly" matching Istio, the bug was copied from `net-istio`, which supplied `retriable-status-codes`, but not a list of codes that are actually retriable!  Looking at the actual Istio code, it seems to be retrying 503s using this mechanism: https://github.com/istio/istio/blob/748c6209f606fb2276bdd4abaf277d4979d172bc/pilot/pkg/networking/core/v1alpha3/route/retry/retry.go#L39

I suspect that this is the remaining delta to resolve the connection reset errors that pop up intermittently with net-contour, but not net-istio...

Fixes: https://github.com/knative-sandbox/net-contour/issues/237

### DIFF
--- a/pkg/reconciler/contour/resources/httpproxy.go
+++ b/pkg/reconciler/contour/resources/httpproxy.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"crypto/sha1"
 	"fmt"
+	"net/http"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


n3wscott (commented):
/lgtm
/approve

Issue #214, opened 2020-08-16
TestServiceCreateAndUpdate 503


mattmoor:
Continuation of #207 and #202.

Failed run: https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-contour-latest/1294948009766817792

mattmoor:
Failed probes:
```
    prober.go:166: "http://service-create-and-update-usqjtxcc.serving-tests.example.com" status = 503, want: 200
    prober.go:167: response: status: 503, body: , headers: map[Content-Length:[0] Date:[Sun, 16 Aug 2020 11:02:25 GMT] Server:[envoy] Zipkin_trace_id:[4b1707a17bebaa33560ef534d2456220]]
    prober.go:166: "http://service-create-and-update-usqjtxcc.serving-tests.example.com" status = 503, want: 200
    prober.go:167: response: status: 503, body: , headers: map[Content-Length:[0] Date:[Sun, 16 Aug 2020 11:02:25 GMT] Server:[envoy] Zipkin_trace_id:[fd28d4babb9573255bc1979f69d1ae06]]
```

mattmoor:
Here are the Envoy logs, with a tight window of 503s among 200s for the same Envoy pod:

```
contour-external/envoy-2qphz[envoy]: [2020-08-16T11:02:25.650Z] "GET / HTTP/1.1" 200 - 0 73 2 2 "35.232.8.97" "Go-http-client/1.1" "eb4d354f-bd8d-445e-80c4-814c385c686b" "service-create-and-update-usqjtxcc.serving-tests.example.com" "10.2
8.6.9:8012"
contour-external/envoy-2qphz[envoy]: [2020-08-16T11:02:25.666Z] "GET / HTTP/1.1" 200 - 0 73 2 2 "35.232.8.97" "Go-http-client/1.1" "eda0759f-a305-47dc-8f55-56af2f83152c" "service-create-and-update-usqjtxcc.serving-tests.example.com" "10.2
8.8.4:8012"
contour-external/envoy-2qphz[envoy]: [2020-08-16T11:02:25.677Z] "GET / HTTP/1.1" 503 NR 0 0 0 - "35.232.8.97" "Go-http-client/1.1" "b740ba70-568b-4000-b8a4-6abca9c84060" "service-create-and-update-usqjtxcc.serving-tests.example.com" "-"
[...]

PR #247, opened 2020-09-08
Ensure that endpoint probes avoid hostname collisions.

mattmoor:
While investigating https://github.com/knative-sandbox/net-contour/issues/214 I noticed that the prober was seeing mismatched hashes when probing the hostnames of the endpoint probe kingress.  I might expect this if the kingress probe were updated, but I believe it was created fresh, so collisions are bizarre.  One plausible explanation of the 503s in the linked issue would be:

1. Main `kingress` reached steady state, endpoint probe `kingress` is deleted.
2. Main `kingress` is reprogrammed.
3. Fresh endpoint probe `kingress` is created with the relevant services.
4. HTTP Proxy from prior endpoint probe is leftover with the exitting service name, and the new endpoint probe updates it instead of creating a new HTTP Proxy.
5. Endpoint probe `kingress` successfully probes the updated HTTP Proxy that it does not own, and reports success.
6. Kubernetes GC catches up and deletes the HTTP Proxy owned by the prior endpoint probe.
7. Thinking the endpoints are adequately covered by the endpoint probe, the main kingress updates the main HTTP Proxy resources.
8. All references to the exitting service have been dropped from HTTP Proxy resources, so the cluster is removed from Envoy.
9. However, the "Cluster" corresponding to that service was still referenced by a "Route", so traffic hitting that "Route" returns a 503.

Reusing HTTP Proxy resources across generations of the kingress probe was never a requirement or goal, so this change simply avoids the possibility of collisions by incorporating the generation of the main `kingress` being probed into the hostnames using when constructing the endpoint probe.

Related: https://github.com/knative-sandbox/net-contour/issues/214

### DIFF
--- a/pkg/reconciler/contour/contour_test.go
+++ b/pkg/reconciler/contour/contour_test.go
@@ -256,7 +256,7 @@ func TestReconcile(t *testing.T) {
 		Key:  "ns/name",
 		Objects: append(append([]runtime.Object{
 			ing("name", "ns", withContour, withGeneration(1), withBasicSpec2),
-			mustMakeProbe(t, ing("name", "ns", withBasicSpec2, withContour), makeItReady),
+			mustMakeProbe(t, ing("name", "ns", withBasicSpec2, withContour, withGeneration(1)), makeItReady),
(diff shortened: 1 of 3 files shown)


tcnghia (approved):
/lgtm

PR #248, opened 2020-09-09
Test linting setup.

markusthoemmes:
title.


markusthoemmes:
Worked as intended. Didn't run without the config file, kicked in with it :+1:

Issue #275, opened 2020-09-28
Ksvc does not become Ready when "previous" HTTProxy's service was garbage collected


nak3:
### Description

If Revision and service were cleaned up by GC before probing the endpoint, Ksvc never becomes Ready and prober keeps trying to probe forever.

#### 0. Configure config-gc immediately collect any inactive revision

```
$ kubectl edit cm -n knative-serving config-g
  ...

  max-non-active-revisions: "0"
  min-non-active-revisions: "0"
  retain-since-create-time: 0s
  retain-since-last-active-time: 0s
```

### 1. Create any Ksvc

e.g.

```
$ kubectl create ns bug
$ kn service create hello-example --image=gcr.io/knative-samples/helloworld-go -n bug
```

#### 2. Update ksvc "twice" immediately

(If you are lucky (=GC was faster than probe), you can produce the error in the first update. But you can produce the error 100% by updating ksvc twice immediately.)

```
$ kn service update --limits-cpu 250m hello-example -n bug
```

NOTE: Update again before ksvc became "Ready". 

```
$ kn service update --limits-cpu 300m hello-example -n bug
```

#### 3. Verify Ksvc status

Now ksvc becomes `EndpointsNotReady` and never become `Ready`.

```
$ kubectl get ksvc -n bug
NAME            URL                                    LATESTCREATED           LATESTREADY             READY     REASON
hello-example   http://hello-example.bug.example.com   hello-example-wfysp-3   hello-example-wfysp-3   Unknown   EndpointsNotReady
```

This is because prober still tries to probe old service in the invalid `HTTPProxy` which was garbage collected. 

```
$ kubectl get httpproxies.projectcontour.io -n bug
NAME                                                              FQDN                                                                TLS SECRET   STATUS    STATUS DESCRIPTION
hello-example--ep-contour-exterc6f42f637b72a763ed9058eb89317401   hello-example-kxfnt-2.gen-2.hello-example.bug.net-contour.invalid                invalid   Spec.Routes unresolved service reference: service "bug/hello-example-kxfnt-2" not found
hello-example--ep-contour-exterfe9da32b241da2e153e3f531dc8b1ef9   hello-example-qbmjv-1.gen-2.hello-example.bug.net-contour.invalid                invalid   Spec.Routes unresolved service reference: service "bug/hello-example-qbmjv-1" not found
[...]

PR #276, opened 2020-09-28
Skip probe previous endpoints when HTTPProxy is not valid

nak3:
If Revision was cleaned up by GC before probing the endpoint, Ksvc
never becomes Ready and prober keeps trying to probe forever.

Please see https://github.com/knative-sandbox/net-contour/issues/275 for the detail.

This patch fixes it by skipping probe if HTTPProxy is not valid.

Fixes https://github.com/knative-sandbox/net-contour/issues/275 https://github.com/knative/serving/issues/9582

/cc @mattmoor @whaught

### DIFF
--- a/pkg/reconciler/contour/contour.go
+++ b/pkg/reconciler/contour/contour.go
@@ -223,7 +223,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) r
 			return err
 		}
 		if diff, err := kmp.SafeDiff(update, matches[0]); err == nil {
-			logger.Debugf("Updated http proxy diff: %s", diff)
+			logger.Debug("Updated http proxy diff: ", diff)
 		} else {
 			logger.Warnw("Error diffing http proxy", zap.Error(err))
 		}
--- a/pkg/reconciler/contour/contour.go
+++ b/pkg/reconciler/contour/contour.go
@@ -223,10 +223,11 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) r
 			return err
 		}
 		if diff, err := kmp.SafeDiff(update, matches[0]); err == nil {
-			logger.Debugf("Updated http proxy: %s", diff)
+			logger.Debugf("Updated http proxy diff: %s", diff)
 		} else {
 			logger.Warnw("Error diffing http proxy", zap.Error(err))
 		}
+		logger.Debugf("Updated http proxy: %#v", update)
 	}
 
 	// Before deleting old programming, check our cache to see whether there is anything to clean up.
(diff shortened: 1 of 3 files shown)


mattmoor:
Given that this is the first we'll be relying on Contour's status, I'm uncomfortable landing this the day before the release 😅 

Let's hold this and land it post-cut Tuesday, so we have a full cycle to bake it.

/hold

pkg/reconciler/contour/resources/kingress.go
```diff
@@ -51,6 +51,12 @@ func MakeEndpointProbeIngress(ctx context.Context, ing *v1alpha1.Ingress, previo
 
 	// Reverse engineer our previous state from the prior generation's HTTP Proxy resources.
 	for _, proxy := range previousState {
+		// Skip probe when status is not valid. It happens when the previous revision was garbage collected.
+		// see: https://github.com/knative/serving/issues/9582
+		if proxy.Status.CurrentStatus != "valid" {
+			continue
+		}
+
```
mattmoor, line 59: I don't think we log the proxy status anywhere (and really don't key off of this anywhere else).  I'd definitely log here (at INFO), but it'd be useful to get logging around the proxy status in generate (at DEBUG).  That would probably help track this and other issues down via logstream.

/unhold

nak3: Thank you. I added Info log. But sorry I didn't understand about where is the best place for `it'd be useful to get logging around the proxy status in generate (at DEBUG)`.

mattmoor: Sorry, I meant that we should put that Info log statement alongside the `continue`.

For the Debug log statement, maybe we should put it in the create/update loop (after it fetches a proxy)?

nak3: Alright. I updated.
The create loop already had the debug log https://github.com/knative-sandbox/net-contour/blob/600a323c5b92d260dcd637058bea1fd3c7f9e15a/pkg/reconciler/contour/contour.go#L211 so just added to update loop.

pkg/reconciler/contour/contour.go
```diff
@@ -223,10 +223,11 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) r
 			return err
 		}
 		if diff, err := kmp.SafeDiff(update, matches[0]); err == nil {
-			logger.Debugf("Updated http proxy: %s", diff)
+			logger.Debug("Updated http proxy diff: ", diff)
```
nak3, line 226: FYI 
L#226 and L#230 prints like logs this:
https://gist.githubusercontent.com/nak3/ebee40ab5c35a25abf4ea87982c0c3bd/raw/642b22546d5227ee2f527542cedd315d0f365d87/contour-httpproxy-example.log

mattmoor (commented):
/lgtm
/approve

PR #278, opened 2020-09-29
Start testing with multi-arch

mattmoor:
Once https://github.com/GoogleContainerTools/distroless/pull/591 lands, this will have us start building and producing multi-arch images.  Only the amd64 flavor will be tested, but this will start to get some mileage through the manifest list path.


tcnghia (approved):
/lgtm

PR #285, opened 2020-09-30
Publish multi-arch during CI

mattmoor:
This still won't work anywhere but amd64 because Envoy is an amd64 image, but 1.16 should start shipping multi-arch images.


tcnghia (approved):
/lgtm

PR #292, opened 2020-10-04
Replace nobody with nonroot

mattmoor:
I'm puzzled that this didn't break in CI, maybe it needs PSPs...

cc @sbawaska fixed this in mink here: https://github.com/mattmoor/mink/pull/73


dprotaso:
upstream fix: 
https://github.com/projectcontour/contour/pull/2985

PR #296, opened 2020-10-05
Use default leader election config

scothis:
Restore default leader election config to avoid contention on the
`contour` ConfigMap. Contour will create a `leader-elect` ConfigMap for
each namespace that is dedicated to the leader election.

This change should allow tools like kapp to update the contour config
with an optimistic lock based on the current resource version.
Previously, the resource would mutate before kapp could complete the
update.

Signed-off-by: Scott Andrews <andrewssc@vmware.com>


mattmoor:
/lgtm
/approve

Thanks!

PR #309, opened 2020-10-11
Clean up cluster suffix a bit.

mattmoor:
We can set this up directly in the job's `env:` using action variables and interpolation.


tcnghia (approved):
/lgtm

PR #322, opened 2020-10-21
Drop --platform=all from upload-test-images.sh

mattmoor:
This runs on every build, so it's unneeded latency, and I think we've gotten the mileage through multi-arch that I wanted.

Especially since distroless now has 4 architectures 🤩


tcnghia (approved):
/lgtm

PR #342, opened 2020-10-28
Narrow down security context of the controller

markusthoemmes:
As per title. The controller should be happy with no capabilities, a read-only filesystem and not being root.


mattmoor (approved):
/lgtm

PR #344, opened 2020-10-30
Enable alpha and beta flag on KinD e2e

nak3:
As per title, this patch enables alpha and beta flag on KinD e2e.

/cc @mattmoor @dprotaso


mattmoor (commented):
/ok-to-test
/lgtm
/approve

PR #406, opened 2021-01-22
Add a missing godoc comment

mattmoor:


n3wscott (approved):

pkg/reconciler/contour/resources/kingress.go
```diff
@@ -31,6 +31,10 @@ import (
 	"knative.dev/pkg/logging"
 )
 
+// MakeEndpointProbeIngress creates a new child kingress resource with a
```
tcnghia, line 34:
```suggestion
// MakeEndpointProbeIngress creates a new child KIngress resource with a
```

mattmoor: I prefer `kingress` because it matches the kubectl short name 🙃

tcnghia (approved):

PR #418, opened 2021-02-08
Add retry test image and remove default retry policy

nak3:
This patch adds retry test image which was added by
https://github.com/knative/networking/commit/45b7ed13aeaa6276d6399cec0e3354c8acd8dbd0.

/kind cleanup

**Release Note**

```release-note
Removed retry policy which retries when backend returns 503.
```

### DIFF
--- a/pkg/reconciler/contour/resources/httpproxy.go
+++ b/pkg/reconciler/contour/resources/httpproxy.go
@@ -21,7 +21,6 @@ import (
 	// nolint:gosec // No strong cryptography needed.
 	"crypto/sha1"
 	"fmt"
-	"net/http"
 	"sort"
 	"strings"
 
@@ -92,9 +91,6 @@ func defaultRetryPolicy() *v1.RetryPolicy {
 			// also retry connection resets.
 			"reset",
 		},
-		RetriableStatusCodes: []uint32{
-			http.StatusServiceUnavailable,
-		},
 	}
 }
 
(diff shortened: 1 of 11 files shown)


nak3:
https://github.com/knative-sandbox/net-contour/pull/246 added Retry policy even after https://github.com/knative/serving/pull/7842 :thinking:

nak3:
/retest

Passed all conformance test without hitting https://github.com/knative-sandbox/net-contour/issues/237 even removing retry setting.

nak3:
/assign @mattmoor 
It seems e2e passes without retry policy now. Is it alight to drop the policy to align with other Ingresses?

pkg/reconciler/contour/resources/httpproxy.go
```diff
@@ -129,22 +108,6 @@ func MakeHTTPProxies(ctx context.Context, ing *v1alpha1.Ingress, serviceToProtoc
...
-			retry := defaultRetryPolicy()
-			if path.DeprecatedRetries != nil && path.DeprecatedRetries.Attempts > 0 {
-				retry.NumRetries = int64(path.DeprecatedRetries.Attempts)
-
-				// When retries is specified explicitly, then we retry some http-level failures as well.
-				retry.RetryOn = append(retry.RetryOn, "5xx")
-
-				if path.DeprecatedRetries.PerTryTimeout != nil {
-					retry.PerTryTimeout = path.DeprecatedRetries.PerTryTimeout.Duration.String()
-				}
```
mattmoor, line 145: By dropping the default you are seriously weakening things, so I'd like to better understand what you are after.

markusthoemmes: I'd think retrying connection level stuff seems fine, so should we drop only the 5xx retries?

@mattmoor for context: @duglin hit an issue where Istio was still exercising retries, though we thought that we didn't use retries at all anymore in Knative.

duglin: As long as the retires don't result in duplicate events being delivered (attempted from the ksvc point of view) then retrying connection errors (but not http errors) might be ok.

mattmoor: the default retry policy contour uses is an extension of the retry policy Istio uses when nothing is specified.  5xx was something we added, which is what the `DeprecatedRetries` field is about, so if this is about dropping that field then 👍 , but some of the code here needs to stay.

nak3: Thank you. I updated as:
- Drop `DeprecatedRetries` handling.
- Drop retry for 5xx from defaultRetryPolicy but keep retry for the connection error.

mattmoor (approved):

PR #421, opened 2021-02-11
Use sigs.k8s.io/yaml for YAML operations

markusthoemmes:
As per title, I recently went through and cleaned a lot of repos where we've been using different YAML libs, sometimes even 2 or 3 in the same repo.

It seemed the best to just align all to the k8s YAML lib, since we're mostly working with YAML in a k8s context anyway.

/assign @mattmoor


tcnghia (approved):
/lgtm

PR #444, opened 2021-03-08
use tmpfs for etcd storage to avoid timeouts

dprotaso:
/kind cleanup
/hold


mattmoor (approved):
/lgtm

If this works well, we should do this everywhere.  cc @n3wscott for `actions-kind` 😉

tcnghia (approved):
/lgtm

PR #449, opened 2021-03-10
Clarify that rule paths should be prefix matched

impl:
Per https://github.com/knative/networking/pull/364, rule paths should be prefix matches and not regular expressions, so the current behavior used here is fine. This PR just updates the doc comment so that there's no confusion about the implementation choice.


tcnghia (approved):
/lgtm

Issue #433, opened 2021-02-19
Take advantage of new PKG helper after release cut


mattmoor:
See: https://github.com/knative/pkg/pull/2032#issuecomment-782189445

PR #454, opened 2021-03-13
Use new filter function constructors

nak3:
This patch changes to use new filter function instroduced by
https://github.com/knative/pkg/pull/2032.

Fix https://github.com/knative-sandbox/net-contour/issues/433

/kind cleanup


tcnghia (approved):
/lgtm

PR #490, opened 2021-05-06
Change certgen imagePullPolicy to IfNotPresent

benmoss:
# Changes

- :broom: Switch certgen job to use `imagePullPolicy: IfNotPresent` instead of `Always`

Upstream they use a mutable :latest tag, but since we are using ko it
will be fully resolved.

https://github.com/projectcontour/contour/pull/2545

This is sorta a bug, if you try to use `kind.local` as your `KO_DOCKER_REPO` you'll get image pull errors for this job.

/kind enhancement

### DIFF
--- a/config/contour/external.yaml
+++ b/config/contour/external.yaml
@@ -1812,7 +1812,7 @@ spec:
       containers:
         - name: contour
           image: ko://github.com/projectcontour/contour/cmd/contour
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
           command:
             - contour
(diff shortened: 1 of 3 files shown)


dprotaso:
/lgtm
/approve

What platform are you running your kind cluster on?

dprotaso (approved):

benmoss:
> What platform are you running your kind cluster on?

This is on linux, why?

dprotaso:
more curious because I've only played with kind on Mac

PR #496, opened 2021-05-13
rename ClusterRole to differ from existing contour installation

### LINKED ISSUE #495: contour ClusterRole clashes with existing contour resource when using kapp
[ClusterRole contour](https://github.com/knative-sandbox/net-contour/blob/ab2f0166b5d80ae220dae1eec9572defe1884e28/config/contour/external.yaml#L1839) will clash with another Contour installation when using tools like `kapp`. Renaming it alongside with ClusterRoleBindings that reference it will prevent problems with existing non-namespaced resources.

```yaml
# line 1
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: contour-external
  labels:
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: contour
subjects:
- kind: ServiceAccount
  name: contour
  namespace: contour-external
---
# line 1839
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  creationTimestamp: null
  name: contour
  labels:
rules:
  - apiGroups:
      - ""
    resources:
      - configmaps
    verbs:
      - create
      - get
      - update
 ...     
```

izabelacg:
# Changes

This PR renames a ClusterRole currently called `contour` and updates its references in ClusterRoleBindings. It also renames the two ClusterRoleBindings that reference the renamed ClusterRole, mainly to avoid collateral effects in case you deploy Knative (version containing the ClusterRole renaming change) over an older version. You would get an error due to changes in an immutable field (roleRef).

The motivation for this PR is to minimize changes necessary to have Knative/net-contour/contour alongside with another contour installation. Currently, if a tool like `kapp` is used to install Knative or Contour, you will get an ownership error because another app already owns a ClusterRole called `contour`.

:broom: Update or clean up current behavior

/kind enhancement

Fixes #495

### DIFF
--- a/config/contour/external.yaml
+++ b/config/contour/external.yaml
@@ -1,13 +1,13 @@
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
-  name: contour-external
+  name: contour-external-crb
   labels:
     networking.knative.dev/ingress-provider: contour
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: contour
+  name: contour-knative
 subjects:
 - kind: ServiceAccount
   name: contour
@@ -1840,7 +1840,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   creationTimestamp: null
-  name: contour
+  name: contour-knative
   labels:
     networking.knative.dev/ingress-provider: contour
 rules:
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -22,6 +22,8 @@ source $(dirname "$0")/../vendor/knative.dev/hack/library.sh
 
 CONTOUR_VERSION="v1.14.0" # This is for controlling which version of contour we want to use.
 
+CLUSTER_ROLE_NAME=contour-knative
+
 FLOATING_DEPS=(
   "github.com/projectcontour/contour@${CONTOUR_VERSION}"
 )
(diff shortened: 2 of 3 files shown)


dprotaso:
you'll have to run `hack/update-deps.sh` and commit the changes

Are there any issues applying this change to an existing installation (ie. upgrading)

hack/update-deps.sh
```diff
@@ -43,6 +45,10 @@ function delete_contour_cluster_role_bindings() {
   sed -e '/apiVersion: rbac.authorization.k8s.io/{' -e ':a' -e '${' -e 'p' -e 'd'  -e '}' -e 'N' -e '/---/!ba' -e '/kind: ClusterRoleBinding/d' -e '}'
 }
 
+function rename_cluster_role() {
+  sed -e "/apiVersion: rbac.authorization.k8s.io/{N;/kind: ClusterRole\b/{N;N;N;s/name: contour/name: $1/}}"
```
dprotaso, line 49: I think after this PR we should replace sed with `ytt` - these expressions are now beyond me lol

izabelacg: `ytt` would be awesome! 😬

config/contour/external.yaml
```diff
@@ -1,13 +1,13 @@
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
-  name: contour-external
+  name: contour-external-crb
```
nak3, line 4: It is better to rename `contour-external-knative` and `contour-internal-knative` for  `ClusterRoleBindings`? 
It will be consistent with ClusterRole name.

hack/update-deps.sh
```diff
@@ -22,6 +22,8 @@ source $(dirname "$0")/../vendor/knative.dev/hack/library.sh
 
 CONTOUR_VERSION="v1.14.0" # This is for controlling which version of contour we want to use.
 
+CLUSTER_ROLE_NAME=contour-knative
```
nak3, line 25: Probably `knative-contour` is better as we have another cluster role `knative-`  prefix?

https://github.com/knative-sandbox/net-contour/blob/12c45bb705cb779b8bb6b4ee73faf36d523c537d/config/200-clusterrole.yaml#L6

izabelacg:
@nak3 I did not mean to close the PR... 😞  is it possible to re-open? Or should I create a new one instead? I put the branch back with the changes you suggested

nak3:
@izabelacg It seems no possible :sweat_smile:  Could you please create a new one?

izabelacg:
oh no, sorry about that 😅 . I created this new one: https://github.com/knative-sandbox/net-contour/pull/500

Issue #385, opened 2020-12-16
Investigate whether contour-operator helps with deploying Contour


jpeach:
The https://github.com/projectcontour/contour-operator repo has a Contour operator that can deploy instances of Contour in different configurations. It might be worth taking a look at whether this can be used for deploying Contour in a Knative configuration

jpeach:
cc @daneyon

nak3:
We should start using operator now as it manages Gateway and GatewayClass on gateway API (Ingress v2).

https://projectcontour.io/guides/gateway-api/

> The preferred method for using Gateway API is with Contour Operator. This is because the GatewayClass and Gateway resources should be managed by active controllers that are implemented by the operator.

nak3:
Knative uses two separated ingress internal and external. And it seems there is no supported way to specify `--ingress-class-name` for operator. 
Luckily https://github.com/projectcontour/contour-operator/issues/228 was opened just today.

PR #499, opened 2021-05-16
Add Contour installed by Operator to kind-e2e test

nak3:
This patch adds contour test installed by operator in kind-e2e test.

net-contour should verify that it works with contour-operator.
Note operator yaml (`./config/contour-operator/`) is not included in release bundle for now.

Fixes https://github.com/knative-sandbox/net-contour/issues/385


hack/tools.go
```diff
@@ -37,5 +37,6 @@ import (
 	_ "knative.dev/networking/config"
 
 	_ "github.com/mikefarah/yq/v3"
+	_ "github.com/projectcontour/contour-operator/cmd"
```
dprotaso, line 40: Do you think we should still be vendoring and building ourselves?

dprotaso:
If so then the operator should probably use the vendored contour

hack/tools.go
dprotaso: If so then the operator should probably use the vendored contour

nak3: I wondered that we should still be building it or not. I think we should not build operator by ourselvs.
Thank you, updated.

dprotaso:
Should we have the operator use our vendored ko version
https://github.com/projectcontour/contour-operator/blob/main/cmd/contour-operator.go#L34-L35

ie. add the flag to on the operator deployment? `--contour-image ko://github.com/projectcontour/contour/cmd/contour`

https://github.com/knative-sandbox/net-contour/blob/eccaac502cac9168c94027a3595b8ad9db26ad23/hack/update-deps.sh#L61-L63

nak3:
Hmm... `ko apply`  command can resolve the image in `--contour-image ko://github.com/projectcontour/contour/cmd/contour` option? I tried but it seems not possible.

dprotaso:
It should work - `ko` checks every string in a yaml file for a `ko://` prefix I believe 

ie. https://github.com/knative/serving/blob/main/config/core/configmaps/deployment.yaml#L27

You would just need to manipulate the operator yaml and add that argument to the container

nak3:
Umm... The configmap works as:

```
$ ko resolve -f config/core/configmaps/deployment.yaml |grep queueSidecarImage
  queueSidecarImage: gcr.io/gcp-compute-engine-223401/queue-39be6f1d08a095bd076a71d288d295b6@sha256:136c0d19dc4094f326b300a20e9737ff0fe8da6b50e1622daa63671469050be1
```

But opertor deployment's command line is not working.

```
$ ko resolve  -f config/contour-operator/operator.yaml  |grep contour-image
            - --contour-image ko://github.com/projectcontour/contour/cmd/contour
```

nak3:
Okay, it seems that the `ko` with the following format can work but it does not work with  `--contour-image` in front of `ko://`.
```yaml
somekey: ko://github.com/projectcontour/contour/cmd/contour
```

dprotaso:
Ah - maybe make it two separate items in the args array?

nak3:
Yeah, the separated line as below works with `ko` but now it hits contour-operator's issue https://github.com/projectcontour/contour-operator/issues/363

```
        - --contour-image 
        - ko://github.com/projectcontour/contour/cmd/contour
```

In short, contour-operator creates label from image tag, but `ko` command creates 64 length tag which exceeds labels length limiation.

dprotaso (approved):
sounds good - let's revisit when they fix the issue upstream

PR #505, opened 2021-05-24
Stop setting DeprecatedVisibility to child ingress

nak3:
This patch removes `DeprecatedVisibility` from child ingress

`DeprecatedVisibility` is deprecated for a long time and actual
visibility is correctly set as:

https://github.com/knative-sandbox/net-contour/blob/2555e84b6b2f39ef524e6a98e1ec03f87ced9c46/pkg/reconciler/contour/resources/kingress.go#L116

Hence this patch stops setting it.

/cc @dprotaso


nak3:
Friendly ping @dprotaso

Issue #409, opened 2021-01-28
Consider renaming `contour-ingress-controller` to something else


dprotaso:
Our deployment name leads people to believe that contour is installed when it is not

https://github.com/knative-sandbox/net-contour/blob/7bff03576e1c01114d47c0fdc58dac74a98115db/config/controller.yaml#L16-L18

dprotaso:
/lifecycle frozen

there seems to be interest to be consistent with other `net-` implementations

https://knative.slack.com/archives/CA9RHBGJX/p1621519009001800

PR #526, opened 2021-06-11
rename controller to be more consistent

dprotaso:
Fixes: https://github.com/knative-sandbox/net-contour/issues/409
Part of https://github.com/knative/networking/issues/448

### DIFF
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -3,7 +3,7 @@
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: knative-contour-core
+  name: knative-serving-contour
   labels:
     networking.knative.dev/ingress-provider: contour
     serving.knative.dev/controller: "true"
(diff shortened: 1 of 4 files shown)


dprotaso:
/hold 

curious if we should use `net-` or `networking-`

config/200-clusterrole.yaml
```diff
@@ -3,7 +3,7 @@
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: knative-contour-core
+  name: knative-serving-contour
```
nak3, line 6: Although `knative-serving-contour` is fine but this is not `knative-net-contour` or `net-contour-core`?

dprotaso: My thinking was 

1) This is consistent with other ingresses
2)  I think it's important to have `serving` in the cluster role name since it's contributing to a `serving` specific aggregated role via the label `serving.knative.dev/controller: "true"`

nak3 (approved):
/lgtm

PR #527, opened 2021-06-11
fix mode of go-licenses vendored code

dprotaso:


nak3:
The same issue happens on net-istio https://github.com/knative-sandbox/net-istio/pull/658/checks?check_run_id=2806525173

But why do we need to change `755` permission? :cold_sweat:  It is so strange to me that we need to have these files the permission.

dprotaso:
@nak3 yeah I changed the tool upstream so we could avoid the hack workaround which I'm removing here: https://github.com/knative/hack/pull/75

dprotaso:
/hold

will wait for `hack` PR to merge

dprotaso:
closing automation will take care of this

PR #540, opened 2021-06-29
Move from using sed to use ytt in the update-deps script

izabelacg:
# Changes

- :broom: Replace use of `sed` with `ytt` for generating external.yaml and internal.yaml. As a result of this PR, very little should change in those two files, basically:
  - two blocks of comments were removed (not sure if they have to be kept or not, so I did not attempt to keep them using ytt)
  - the rest is just some indentation changes done by ytt. This tool helps comparing the files ignoring the whitespaces changes: https://quickdiff.net/
  - ~~**WIP**: actually there is this [change here](https://github.com/knative-sandbox/net-contour/pull/540/files#diff-78c86fc3794c9c7594e32c834b7985d825248d9a1b4dd1d88da0bde328ebc0ebR2088) that I don't understand where is coming from: defaultMode 420 instead of 0644.~~

/kind <kind>


hack/overlays/values.yaml
```diff
@@ -0,0 +1,5 @@
+#@data/values
+---
+namespace:
+clusterrole:
+  name:
```
dprotaso, line 5: do we need to declare this file if we're not setting any default values?

hack/overlays/overlay-envoy.yaml
```diff
@@ -0,0 +1,54 @@
...
+---
+spec:
+  #@overlay/remove
+  externalTrafficPolicy: Local
+  #@overlay/match
+  type: ClusterIP
+#@ end
+
+#@overlay/match by=overlay.subset({"kind":"DaemonSet","metadata":{"name": "envoy"}})
+#@overlay/match-child-defaults when=1
```
dprotaso, line 17: Since the match above doesn't have an `expects` set I'm assuming it defaults to `1`

Does that mean `when=1` here is redundant?

hack/overlays/overlay-envoy.yaml
```diff
@@ -0,0 +1,54 @@
...
+            - containerPort: 8080
+              #@overlay/remove
+              hostPort: 80
+            #@overlay/match by="containerPort"
+            - containerPort: 8443
+              #@overlay/remove
+              hostPort: 443
+
+#@overlay/match by=overlay.subset({"kind":"DaemonSet","metadata":{"name": "envoy"}})
+#@overlay/match-child-defaults when=1
```
dprotaso, line 36: Same comment as above

hack/overlays/overlay-image.yaml
```diff
@@ -0,0 +1,34 @@
...
+      containers:
+        #@overlay/match by="name"
+        - name: contour
+          #@overlay/replace via=lambda left, right: left.replace(left, ko_import_path)
+          image:
+          imagePullPolicy: IfNotPresent
+        #@overlay/match by="name"
+        - name: shutdown-manager
+          #@overlay/replace via=lambda left, right: left.replace(left, ko_import_path)
+          image:
```
dprotaso, line 21: Is it possible to just have 

```suggestion
          image: ko://github.com/projectcontour/contour/cmd/contour
```

like you do lower in the file?

hack/overlays/overlay-image.yaml
```diff
@@ -0,0 +1,34 @@
+#@ load("@ytt:overlay", "overlay")
+#! rewrite image and image pull policy
+
+#@ ko_import_path = "ko://github.com/projectcontour/contour/cmd/contour"
+
+#@overlay/match by=overlay.all,expects=17
+#@overlay/match-child-defaults when=1
```
dprotaso, line 7: I think we should only have `expects="1+"` otherwise I can see this being fairly brittle

or we target specific deployments/daemonsets

hack/overlays/overlay-image.yaml
```diff
@@ -0,0 +1,34 @@
...
+          image:
+          imagePullPolicy: IfNotPresent
+        #@overlay/match by="name"
+        - name: shutdown-manager
+          #@overlay/replace via=lambda left, right: left.replace(left, ko_import_path)
+          image:
+          imagePullPolicy: IfNotPresent
+
+#@overlay/match by=overlay.all,expects=17
+#@overlay/match-child-defaults when=1
```
dprotaso, line 25: same comment as above regarding the hardcoded numbers

hack/update-deps.sh
```diff
@@ -95,63 +53,5 @@ function contour_operator_yaml() {
...
-  | rewrite_contour_namespace contour-external \
-  | rewrite_serve_args contour-external | rewrite_user \
-  | rewrite_image | rewrite_image_pull_policy | rewrite_command | disable_hostport \
-  | add_ingress_provider_labels >> config/contour/external.yaml
-
-rm -rf config/contour-operator/operator.yaml
-
-contour_operator_yaml \
-  >> config/contour-operator/operator.yaml
+contour_yaml | run_ytt --ignore-unknown-comments --data-value namespace=contour-internal --data-value clusterrole.name=$CLUSTER_ROLE_NAME -f hack/overlays -f - >> config/contour/internal.yaml
```
dprotaso, line 56: can we format the args on a new line so it's easier to read the whole command

```suggestion
contour_yaml | \
  run_ytt --ignore-unknown-comments \
    --data-value namespace=contour-internal \
    --data-value clusterrole.name=$CLUSTER_ROLE_NAME \
    -f hack/overlays \
    -f - >> config/contour/internal.yaml
```

dprotaso (commented):
Can we change the overlay file names to be more descriptive - 

ie. 

instead of `overlay-clusterrolebindings.yaml` - we just call it `delete-contour-clusterrolebinding.yaml`


then we can drop the comments if needed

dprotaso:
> two blocks of comments were removed (not sure if they have to be kept or not, so I did not attempt to keep them using ytt)

Weird are other comments retained?

> WIP: actually there is this change here that I don't understand where is coming from: defaultMode 420 instead of 0644.

I think that should be sorted - we had an issue with out go-licenses tool

dprotaso:
> WIP: actually there is this change here that I don't understand where is coming from: defaultMode 420 instead of 0644.

I misread - I think ytt is rewriting 0644 (octal) to 420 (decimal) which is weird but fine

hack/overlays/clusterrolebindings.yaml
```diff
@@ -0,0 +1,19 @@
...
+#@ load("@ytt:data", "data")
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  #@yaml/text-templated-strings
+  name: knative-(@= data.values.namespace @)
+  labels:
+    networking.knative.dev/ingress-provider: contour
```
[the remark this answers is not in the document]
izabelacg: Yes, we can drop it as long as the overlay that adds the label is applied after this one. I'll rename the files and confirm that it is the case.

hack/overlays/values.yaml
izabelacg: I tried to remove the values.yaml file and got an error, I'll take a look at the docs to see if it's required or not

hack/overlays/overlay-envoy.yaml
izabelacg: Your are right, it's redundant :). Will remove it

hack/overlays/overlay-image.yaml
izabelacg: Yes 😂 , I don't know why I was using lamba for this...

izabelacg:
> > two blocks of comments were removed (not sure if they have to be kept or not, so I did not attempt to keep them using ytt)
> 
> Weird are other comments retained? I don't think this is blocking
> 

The block of comments that are not kept:
```yaml
apiVersion: v1
kind: Service
metadata:
 name: envoy
 namespace: contour-external
 annotations:
  # This annotation puts the AWS ELB into "TCP" mode so that it does not
  # do HTTP negotiation for HTTPS connections at the ELB edge.
  # The downside of this is the remote IP address of all connections will
  # appear to be the internal address of the ELB. See docs/proxy-proto.md
  # for information about enabling the PROXY protocol on the ELB to recover
  # the original remote IP address.
  service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
 labels:
...
```

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
 labels:
  app: contour
 name: contour
 namespace: contour-external
spec:
 replicas: 2
 strategy:
  type: RollingUpdate
  rollingUpdate:
   # This value of maxSurge means that during a rolling update
   # the new ReplicaSet will be created first.
   maxSurge: 50%
...
```

This one is in between yaml docs:
```yaml
# This file is generated from the individual YAML files by generate-deployment.sh. Do not
# edit this file directly but instead edit the source files and re-render.
#
# Generated from:
#    examples/contour/00-common.yaml
#    examples/contour/01-contour-config.yaml
#    examples/contour/01-crds.yaml
#    examples/contour/02-job-certgen.yaml
#    examples/contour/02-rbac.yaml
#    examples/contour/02-role-contour.yaml
#    examples/contour/02-service-contour.yaml
#    examples/contour/02-service-envoy.yaml
#    examples/contour/03-contour.yaml
#    examples/contour/03-envoy.yaml
#
```

hack/overlays/clusterrolebindings.yaml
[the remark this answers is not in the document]
izabelacg: > Can we change the overlay file names to be more descriptive -
> 
> ie. instead of `overlay-clusterrolebindings.yaml` - we just call it `delete-contour-clusterrolebinding.yaml` - then we can drop the comments if needed

I just pushed the changes. Let me know of any suggestions for the file names

dprotaso:
@googlebot I consent.

dprotaso:
/lgtm
/approve

@izabelacg looks good - you just need to mark the PR 'ready for review'

dprotaso:
🎉 thanks!

PR #563, opened 2021-07-27
Cherry-pick fix for CVE-2021-32783

evankanderson:
This is a fix for [CVE-2021-32783](https://github.com/projectcontour/contour/security/advisories/GHSA-5ph6-qq5x-7jwc).

Process: I read https://github.com/projectcontour/contour/compare/v1.17.0...v1.17.1 and attempted to apply the patch to the vendored copy of Contour. Unlike the real fix, I didn't provide a configmap value to override to allow ExternalName services, so an upgrade to 1.17.1 or 1.18 will be needed to re-enable the functionality.

(I intend to copy this cherry-pick to release branches, since I'm not sure how easily it can be merged at `main`.)

### DIFF
--- a/vendor/github.com/projectcontour/contour/internal/dag/accessors.go
+++ b/vendor/github.com/projectcontour/contour/internal/dag/accessors.go
@@ -57,6 +57,11 @@ func (dag *DAG) EnsureService(meta types.NamespacedName, port intstr.IntOrString
 		return nil, err
 	}
 
+	err = validateExternalName(svc, false)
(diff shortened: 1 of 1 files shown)


evankanderson:
/assign @nak3 @julz @kvmware 

/hold

I'd like a couple thoughts on whether this is the right way to do this to get releases out.

evankanderson:
/close

It looks like this won't pass the build tests. See #564 for a version that will.

PR #564, opened 2021-07-27
Cherry-pick fix for CVE-2021-32783 via go.mod replace directive

evankanderson:
Replacement for #563.

This is a fix for [CVE-2021-32783](https://github.com/projectcontour/contour/security/advisories/GHSA-5ph6-qq5x-7jwc).

Process: I read https://github.com/projectcontour/contour/compare/v1.17.0...v1.17.1 and attempted to apply the patch to the vendored copy of Contour. Unlike the real fix, I didn't provide a configmap value to override to allow ExternalName services, so an upgrade to 1.17.1 or 1.18 will be needed to re-enable the functionality.

https://github.com/evankanderson/contour/tree/CVE-32783 contains the fix; Contour is only maintaining the latest release branch.

### DIFF
--- a/vendor/github.com/projectcontour/contour/internal/dag/accessors.go
+++ b/vendor/github.com/projectcontour/contour/internal/dag/accessors.go
@@ -57,6 +57,11 @@ func (dag *DAG) EnsureService(meta types.NamespacedName, port intstr.IntOrString
 		return nil, err
 	}
 
+	err = validateExternalName(svc, false)
+	if err != nil {
+		return nil, err
+	}
+
 	if dagSvc := dag.GetService(k8s.NamespacedNameOf(svc), svcPort.Port); dagSvc != nil {
 		return dagSvc, nil
 	}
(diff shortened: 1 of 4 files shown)


evankanderson:
/assign @nak3 @julz @kvmware

This seems "cleaner" in a way, though having a random branch with a commit that's insipired by a later Contour release is... not great.

For 0.25, we should move to 0.17.1 or 0.18.0 (depending on schedule).

evankanderson:
If you like this one, you'll probably like #565 (release 0.23), #566 (release 0.22), and #567 (release 0.21), too. Per https://github.com/knative/community/blob/main/mechanics/RELEASE-VERSIONING-PRINCIPLES.md#knative-community-support-window-principle:

> We (the community) support the most recent 4 versions of Knative. The term support as used here means that we will maintain release branches for these releases. We will consider backporting applicable fixes to these release branches depending on severity and feasibility.

vendor/github.com/projectcontour/contour/internal/dag/accessors.go
```diff
@@ -78,6 +83,38 @@ func (dag *DAG) EnsureService(meta types.NamespacedName, port intstr.IntOrString
...
+func validateExternalName(svc *v1.Service, enableExternalNameSvc bool) error {
+
+	// If this isn't an ExternalName Service, we're all good here.
+	en := externalName(svc)
+	if en == "" {
+		return nil
+	}
+
+	// If ExternalNames are disabled, then we don't want to add this to the DAG.
+	if !enableExternalNameSvc {
```
nak3, line 95: Since DomainMapping's HTTProxy points to the ExternalName service, we need to set `enableExternalNameSvc` to true.
The current failure of prow is due to this.

nak3 (commented):
The alternative workaround is that all services are listed before creating HTTPProx at 
https://github.com/knative-sandbox/net-contour/blob/646c44cf21b7a544446e971a4401aa39c544ab83/pkg/reconciler/contour/contour.go#L184-L187 so we can do same validation with the `validateExternalName` there. But I think this PR's approach is better.

nak3:
> Unlike the real fix, I didn't provide a configmap value to override to allow ExternalName services, so an upgrade to 1.17.1 or 1.18 will be needed to re-enable the functionality.

I think this will be a big pain... We need to ask users to enable [`enableExternalNameService`](https://github.com/projectcontour/contour/blob/f62e3cb3bd9ca69b05c5d292e34b7ef8842ad121/examples/render/contour.yaml#L89-L93) to `true`, which is not recommended by upstream, whenever they want to use Contour with Knative.

vendor/github.com/projectcontour/contour/internal/dag/accessors.go
evankanderson: Ugh, I didn't test that manually.

If we set `enableExternalNameSvc` to true, we're effectively defeating much of the CVE protection. 😢

julz: Just to say this out loud (not suggesting we hack this change in here since we'd be forking contour that way, obviously, but maybe it triggers a better idea in someone - or _maybe_ we can suggest this upstream), a potentially safe thing to do might be to allow targets of the form "name.ns.svc.cluster.local" (ie allowlist rather than denylist). That would - I think - work fine for our use cases and I don't _think_ there's any way those can resolve to localhost since that dns space is managed by k8s.

evankanderson:
Hmm, I'm wondering if we need to change the implementation of DomainMapping to use https://projectcontour.io/docs/v1.17.1/config/inclusion-delegation/

vendor/github.com/projectcontour/contour/internal/dag/accessors.go
dprotaso: > we're effectively defeating much of the CVE protection

Doesn't the `localhostNames` check below resolve that?

dprotaso: Nm - it was pointed out to me you could have a non-obvious ExternalName that resolves to 127.0.0.1

Issue #423, opened 2021-02-11
Probing always fails when HTTP fallback is disabled in config-network


dprotaso:
When HTTP fallback is disabled probing will set an `https` scheme and a port of `443`

https://github.com/knative-sandbox/net-contour/blob/50b263c04bb6ff46398f9a2baa06cd9f13cc0ba3/pkg/reconciler/contour/lister.go#L48-L52

Contour targets Pod IPs for probing which doesn't do TLS termination so the endpoints will never be ready.

nak3:
/assign

I will work on this.

dprotaso:
Note this issue changes slightly when https://github.com/knative-sandbox/net-contour/pull/610 lands

Since we'll want to configure the spec.HTTPOption now

kvmware:
When using `kubectl apply -f https://github.com/knative/net-contour/releases/download/v0.25.0/contour.yaml` and `kubectl apply -f https://github.com/knative/net-contour/releases/download/v0.25.0/net-contour.yaml` a service is created `kn service create hello --image gcr.io/knative-samples/helloworld-go --port 8080 --env TARGET=New_World --revision-name=new_world` which results in:

**`$ kubectl get ksvc`**
```
NAME    URL                                           LATESTCREATED   LATESTREADY   READY     REASON
hello   https://hello.default.34.71.36.176.sslip.io   hello-world     hello-world   Unknown   EndpointsNotReady
```

**`$ kubectl describe ksvc hello`**
```
...
Status:
  Address:
    URL:  http://hello.default.svc.cluster.local
  Conditions:
    Last Transition Time:        2021-09-23T19:05:43Z
    Status:                      True
    Type:                        ConfigurationsReady
    Last Transition Time:        2021-09-23T19:05:44Z
    Message:                     Waiting for Envoys to receive Endpoints data.
    Reason:                      EndpointsNotReady
    Status:                      Unknown
    Type:                        Ready
    Last Transition Time:        2021-09-23T19:05:44Z
    Message:                     Waiting for Envoys to receive Endpoints data.
    Reason:                      EndpointsNotReady
    Status:                      Unknown
    Type:                        RoutesReady
  Latest Created Revision Name:  hello-world
  Latest Ready Revision Name:    hello-world
  Observed Generation:           1
  Traffic:
    Latest Revision:  true
    Percent:          100
    Revision Name:    hello-world
  URL:                https://hello.default.34.71.36.176.sslip.io
...
```

[...]

PR #610, opened 2021-09-08
Support HTTPOption in Ingress spec

nak3:
This patch supports Ingress's `spec.HTTPOption`.

As a consequence, we can drop the dependency of config-network configmap.

Fix https://github.com/knative/networking/issues/303

/cc @dprotaso @kvmware


nak3:
/retest

Not related failure.

kvmware:
LGTM, Would want @dprotaso to do sign off though.

dprotaso:
> As a consequence, we can drop the dependency of config-network configmap.

Is serving now setting the default value for `spec.HTTPOption`  based on the `config-network` setting?

nak3:
Yes, it was done by https://github.com/knative/serving/commit/8179fc67dbb20144265a7ee0f77810b318b2aaca

pkg/reconciler/contour/lister.go
```diff
@@ -44,15 +44,17 @@ func (l *lister) ListProbeTargets(ctx context.Context, ing *v1alpha1.Ingress) ([
...
-	}
-
 	visibilityKeys := cfg.Contour.VisibilityKeys
+
 	for key, hosts := range ingress.HostsPerVisibility(ing, visibilityKeys) {
+		port, scheme := int32(80), "http"
+
+		// Probe external servce with https.
+		if ing.Spec.HTTPOption == v1alpha1.HTTPOptionRedirected &&
+			!visibilityKeys["ClusterLocal"].Has(key) {
```
dprotaso, line 54: cc @kvmware

I wonder if this line fixes this issue: https://github.com/knative-sandbox/net-contour/issues/423

dprotaso (approved):

Issue #631, opened 2021-10-05
Add resource requests/limits to our contour envoy daemonsets


dprotaso:
It looks like our networking conformance tests have been pretty flakey
https://testgrid.k8s.io/r/knative-own-testgrid/net-contour#continuous

Running some tests locally I see the envoy containers restarting around the time 503 errors occur

Note: envoy returns the following status codes 

503/unknown cluster => there's no backing endpoints ([link](https://github.com/envoyproxy/envoy/blob/5b1015cffbf6e02a7ccec9c12a6b7f246f8f3d30/source/common/router/router.cc#L444))
404/no cluster match for url => there's not backing ([link](https://github.com/envoyproxy/envoy/blob/5b1015cffbf6e02a7ccec9c12a6b7f246f8f3d30/source/common/router/router.cc#L391))

The envoy pods restart because liveness probes fail due to timeout. Digging a bit it looks like we don't have resource requests/limits set on our envoy daemon set pods.

We should probably create an ytt overlay to add these requests/limits in order to reserve the CPU:

The overlays are applied here:
https://github.com/knative-sandbox/net-contour/blob/2db64e2d558a32fe550e1f478bf0ce76e98a3673/hack/update-deps.sh#L56-L68

carlisia:
Working on this...

PR #647, opened 2021-10-25
🧹 Add resource requests/limits

carlisia:
# Changes

Fix test failures by increasing resource limits.

/kind performance

Fixes https://github.com/knative-sandbox/net-contour/issues/631

/assign @dprotaso 
/hold

### DIFF
--- a/hack/overlays/set-envoy-daemonset-resources-limits.yaml
+++ b/hack/overlays/set-envoy-daemonset-resources-limits.yaml
@@ -1,9 +1,9 @@
 #@ load("@ytt:overlay", "overlay")
 #@ load("@ytt:data", "data")
 
-#@overlay/match by=overlay.all,expects="1+"
-#@overlay/match-child-defaults when=1
+#@overlay/match by=overlay.subset({"kind":"DaemonSet"}),expects=[1]
 ---
+#@overlay/match-child-defaults missing_ok=True
 spec:
   template:
     spec:
--- a/hack/overlays/set-envoy-daemonset-resources-limits.yaml
+++ b/hack/overlays/set-envoy-daemonset-resources-limits.yaml
@@ -1,0 +1,27 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+
+#@overlay/match by=overlay.subset({"kind":"DaemonSet","metadata":{"name": "envoy"}})
+---
+spec:
+  template:
+    spec:
+      containers:
+        #@overlay/match by="name"
+        - name: shutdown-manager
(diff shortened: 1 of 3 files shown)


hack/overlays/set-envoy-daemonset-resources-limits.yaml
```diff
@@ -0,0 +1,28 @@
...
+#@overlay/match by=overlay.all,expects="1+"
+#@overlay/match-child-defaults when=1
+---
+spec:
+  template:
+    spec:
+      containers:
+        #@overlay/match by="name"
+        - name: shutdown-manager
+          resources:
```
dprotaso, line 13: Dropping the `when=1` shows that the match doesn't happen since the `resources` block is missing from the upstream contour example yaml. So we need to annotate that if it's missing that's ok

```suggestion
          #@overlay/match missing_ok=True
          resources:
```

hack/overlays/set-envoy-daemonset-resources-limits.yaml
```diff
@@ -0,0 +1,28 @@
...
+            requests:
+              cpu: 40m
+              memory: 40Mi
+            limits:
+              cpu: 400m
+              memory: 400Mi
+        #@overlay/match by="name"
+        - name: envoy
+          resources:
```
dprotaso, line 22: We'll want want the same annotation as above

```suggestion
          #@overlay/match missing_ok=True
          resources:
```

hack/overlays/set-envoy-daemonset-resources-limits.yaml
```diff
@@ -0,0 +1,28 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+
+#@overlay/match by=overlay.all,expects="1+"
+#@overlay/match-child-defaults when=1
```
dprotaso, line 5:
```suggestion
#@overlay/match by=overlay.subset({"kind":"DaemonSet"}),expects="1+"
```

We need the subset since the spec below only applies to DaemonSets. 

Also I think the `when=1` is hiding the fact that we aren't matching for some reason so the overlay isn't applied. Typically we should see the result of these overlays in `config/contour/external.yaml` and `config/contour/internal.yaml` after running `./hack/update-deps.sh`

carlisia: Ah...... thanks for pointing this out @dprotaso. I fixed it, now it has applied properly and the config files have the resources added to them.

dprotaso:
🎉 🎉 

/lgtm
/approve

PR #661, opened 2021-11-05
Add k8s recommended labels

psschwei:
# Changes

Adds Kubernetes recommended labels to networking as discussed in https://github.com/knative/networking/issues/552

Note that changes to the labels in `vendor/knative.dev/networking` should be handled by the automation once https://github.com/knative/networking/pull/555 merges. 

/kind enhancement

### DIFF
--- a/config/config-contour.yaml
+++ b/config/config-contour.yaml
@@ -19,6 +19,9 @@ metadata:
   namespace: knative-serving
   labels:
     networking.knative.dev/ingress-provider: contour
+    app.kubernetes.io/name: contour
+    app.kubernetes.io/part-of: knative-serving
+    app.kubernetes.io/version: devel
     serving.knative.dev/release: devel
 data:
   _example: |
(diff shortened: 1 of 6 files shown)


config/config-contour.yaml
```diff
@@ -19,6 +19,9 @@ metadata:
   namespace: knative-serving
   labels:
     networking.knative.dev/ingress-provider: contour
+    app.kubernetes.io/name: contour
```
dprotaso, line 22: Let's use `net-contour` since this is technically the networking contour plugin
```suggestion
    app.kubernetes.io/name: net-contour
```

psschwei: I'm sensing a theme 😄

nak3:
LGTM
/assign @dprotaso

dprotaso:
Can we add the annotations to

`config/200-clusterrole.yaml`
`config/controller.yaml`

psschwei:
done (neither of those had the original devel label, which I'm guessing is why they didn't get the initial update)

dprotaso:
/lgtm
/approve

thanks!

PR #701, opened 2022-01-12
pin deps

dprotaso:
- pin dependencies so we can build knative & contour with the desired libraries
- drop yq from tools since we don't need it

This should address the failures seen here: https://github.com/knative-sandbox/net-contour/pull/697

Those are caused by knative.dev/pkg bumping the min version for a transient dependency that breaks building contour in this repo


dprotaso:
Closing in favour of: https://github.com/knative-sandbox/net-contour/pull/702

Issue #504, opened 2021-05-20
Ensure kind operator e2e test uses the same built images as our other tests


dprotaso:
Yeah, the separated line as below works with `ko` but now it hits contour-operator's issue https://github.com/projectcontour/contour-operator/issues/363

```
        - --contour-image 
        - ko://github.com/projectcontour/contour/cmd/contour
```

In short, contour-operator creates label from image tag, but `ko` command creates 64 length tag which exceeds labels length limiation.

_Originally posted by @nak3 in https://github.com/knative-sandbox/net-contour/issues/499#issuecomment-844675891_

dprotaso:
While our other kind e2e tests are build from source in our vendor folder the operator currently does not

nak3:
This should be able to fix by Contour 1.16 https://github.com/projectcontour/contour-operator/commit/f19b8b6333dd70b324e0d997ed7ba3c320c9be8c

PR #702, opened 2022-01-12
Re-use upstream contour image for cmd/contour

dprotaso:
Fixes: https://github.com/knative-sandbox/net-contour/issues/504

- drop yq from tools - it's not used
- don't vendor cmd/contour

Motivation: We get into weird dependency issues since upstream contour and net-contour are not aligned. This change un-vendors the contour go binary and uses what's built upstream.

The alternative is: https://github.com/knative-sandbox/net-contour/pull/701

```release-note
We're now shipping contour's upstream images instead of rebuilding our own
```


nak3:
I just wanted to note that some "meta" impact by this change:
- The contour image may not support arch which was supported by `ko --platform=all` but not supported by contour upstream.
- The contour image will use different base image from current distroless image. 
These change does not impact on users :thinking:  Anyway I am personally fine with this change, though.

/assign @carlisia

dprotaso:
Good point we should open an issue regarding supporting multiple archs

dprotaso:
Looks like they support amd64 & arm64
```sh
$ crane manifest ghcr.io/projectcontour/contour:v1.19.1
{
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "schemaVersion": 2,
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "digest": "sha256:c3440419e63ed74851e7d9127b91479f6eac67343015b18e94a61539c523f6a2",
         "size": 529,
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "digest": "sha256:45c9f2b409f8641e6653816158128b39caecd857016156a82c15ffc1147c9a00",
         "size": 529,
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}%
```


We also supported `arm/linux` `ppc64le/linux` and `s390x/linux`. But the envoy image doesn't support these

dprotaso:
> The contour image will use different base image from current distroless image.

They build from scratch so I don't think it's an issue

carlisia:
/lgmt
/approve


2022-01-13 at 10 19 AM

( ͡° ͜ °)

carlisia:
oh lord

PR #706, opened 2022-01-17
drop dead/unused symlink

dprotaso:


nak3:
/retest

https://github.com/knative-sandbox/net-contour/pull/710, which should fix the CI issue, was merged just now.

PR #713, opened 2022-01-21
Copy parent HTTPOption to endpoint probe, if present.

### LINKED ISSUE #712: net-contour fails to reconcile if only the HTTP contour port is available
Debugged this with a user today. The following setup suffices:

1. Shared TLS certificate in `contour-external/shared-wildcard` Secret. I used [cert-manager](https://cert-manager.io/docs/) to do this, but you could use OpenSSL or anything else. I don't think the actual CN/DNS name matters.
2. `kubectl patch -n knative-serving configmap config-contour --type merge --patch '{"data": {"default-tls-secret": "contour-external/shared-wildcard"}}'`
3. `kubectl patch -n knative-serving configmap config-network --type merge --patch '{"data": {"ingress.class": "contour.ingress.networking.knative.dev", "http-protocol": "Redirected", "default-external-scheme": "https"}}'`
4. `kubectl edit -n contour-external service envoy` and remove the port 80 port, leaving only the port 443 port.

(I don't have access to the actual setup, but the short version was that Contour was pre-configured with an HTTPS certificate, and they didn't want any HTTP traffic to be handled at all.)

When you attempt to deploy a new Service, the `kn service create` will time out with a reason `EndpointsNotReady`. The `net-contour-controller` will have log messages like the following:

```
{"severity":"INFO","timestamp":"2022-01-21T07:31:33.8744081Z","logger":"net-contour-controller.event-broadcaster","caller":"record/event.go:282","message":"Event(v1.ObjectReference{Kind:\"Ingress\", Namespace:\"default\", Name:\"hello2--ep\", UID:\"4bc2c217-a110-4bfd-afaf-88c75f53d8a2\", APIVersion:\"networking.internal.knative.dev/v1alpha1\", ResourceVersion:\"66821\", FieldPath:\"\"}): type: 'Warning' reason: 'InternalError' failed to probe Ingress default/hello2--ep: failed to lookup port 80 in contour-external/envoy: no port with number 80 found","commit":"a248e38"}
```

Note that this is attempting to lookup port 80 on the envoys, despite all traffic being routed over 443.

(Manually-tested fix incoming)

evankanderson:
Fixes #712

### DIFF
--- a/pkg/reconciler/contour/resources/httpproxy.go
+++ b/pkg/reconciler/contour/resources/httpproxy.go
@@ -97,11 +97,10 @@ func MakeHTTPProxies(ctx context.Context, ing *v1alpha1.Ingress, serviceToProtoc
 	ing = ing.DeepCopy()
 	ingress.InsertProbe(ing)
 
-	hostToTLS := make(map[string]*v1alpha1.IngressTLS, len(ing.Spec.TLS))
+	hostToTLS := make(map[string]v1alpha1.IngressTLS, len(ing.Spec.TLS))
 	for _, tls := range ing.Spec.TLS {
 		for _, host := range tls.Hosts {
-			t := tls
-			hostToTLS[host] = &t
+			hostToTLS[host] = tls
 		}
 	}
 
--- a/pkg/reconciler/contour/resources/kingress.go
+++ b/pkg/reconciler/contour/resources/kingress.go
@@ -47,12 +47,16 @@ func MakeEndpointProbeIngress(ctx context.Context, ing *v1alpha1.Ingress, previo
 			OwnerReferences: []metav1.OwnerReference{*kmeta.NewControllerRef(ing)},
 		},
 		Spec: v1alpha1.IngressSpec{
-			HTTPOption: ing.Spec.HTTPOption,
+			HTTPOption: v1alpha1.HTTPOptionEnabled,
 		},
 	}
 
-	if childIng.Spec.HTTPOption == "" {
-		childIng.Spec.HTTPOption = v1alpha1.HTTPOptionEnabled
+	hasCert := len(ing.Spec.TLS) > 0 || config.FromContext(ctx).Contour.DefaultTLSSecret != nil
+
+	if ing.Spec.HTTPOption == v1alpha1.HTTPOptionRedirected && hasCert {
+		// Set the probe to operate over HTTPS IFF we have certificates AND are TLS-required
+		childIng.Spec.HTTPOption = v1alpha1.HTTPOptionRedirected
+		childIng.Spec.TLS = append(childIng.Spec.TLS, ing.Spec.TLS...)
 	}
 
 	sns := ServiceNames(ctx, ing)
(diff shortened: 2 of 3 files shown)


evankanderson:
I have no idea how I bollixed up the first commit, corrected.

pkg/reconciler/contour/resources/httpproxy.go
```diff
@@ -97,11 +97,10 @@ func MakeHTTPProxies(ctx context.Context, ing *v1alpha1.Ingress, serviceToProtoc
 	ing = ing.DeepCopy()
 	ingress.InsertProbe(ing)
 
-	hostToTLS := make(map[string]*v1alpha1.IngressTLS, len(ing.Spec.TLS))
+	hostToTLS := make(map[string]v1alpha1.IngressTLS, len(ing.Spec.TLS))
 	for _, tls := range ing.Spec.TLS {
 		for _, host := range tls.Hosts {
-			t := tls
-			hostToTLS[host] = &t
+			hostToTLS[host] = tls
```
evankanderson, line 103: Since we were making a copy, there was no need to use references here anyway.

pkg/reconciler/contour/resources/kingress.go
```diff
@@ -47,8 +47,6 @@ func MakeEndpointProbeIngress(ctx context.Context, ing *v1alpha1.Ingress, previo
 			OwnerReferences: []metav1.OwnerReference{*kmeta.NewControllerRef(ing)},
 		},
 		Spec: v1alpha1.IngressSpec{
-			// TODO: Probing against HTTP should be enough as it ensures Envoy's EDS?
-			// Need to verify it by scale-N test with HTTPS.
```
evankanderson, line 51: Since this converts things to probe HTTPS for HTTPS instances with certificates, I think that should be sufficient for this TODO.

evankanderson (commented):
This is ready for review.

dprotaso:
Since we're opening up the ability to probe HTTPS endpoints we'll probably want the following

1. The ability to included an addition CA cert/intermediate certs that the prober uses
   - This is for handling corporate environments with a non public CA 
   - or we confirm and document `SSL_CERT_FILE`/`SSL_CERT_DIR` works as expected
2. (optional) Configure an option to skip cert verification [`crypto/tls/Config.InsecureSkipVerify `](https://pkg.go.dev/crypto/tls#Config) 
   - Probably include big warnings that this should be off - or use only in testing
3. E2E test converting this functionality

evankanderson:
(This change failed until I included the right hostnames in the child KIngress's `spec.tls` field, so I think we do have e2e test coverage.)

Re item 2, we already configure `InsecureSkipVerify` in the networking prober:

https://github.com/knative/networking/blob/57bee6625a8ac19f6a1f02261d23245facf98ed0/pkg/status/status.go#L368

I'm not sure what you're thinking of for item 1, given item 2.

dprotaso:
> (This change failed until I included the right hostnames in the child KIngress's spec.tls field, so I think we do have e2e test coverage.)

Great

> Re item 2, we already configure InsecureSkipVerify in the networking prober:

Oh interesting - thought it was stricter. Let's forgo 1 &2 then.

PR #722, opened 2022-02-05
Update documentation with e2e testing instructions for kind

evankanderson:
These aren't perfect, but they should at least give a start and a chance for other people to add to them.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -24,10 +24,18 @@ You must install these tools:
 1. [`go`](https://golang.org/doc/install): The language Knative `net-contour` is
    built in
 1. [`git`](https://help.github.com/articles/set-up-git/): For source control
-1. [`dep`](https://github.com/golang/dep): For managing external dependencies.
 1. [`ko`](https://github.com/google/ko): For development.
 1. [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/): For
    managing development environments.
+
+If you want to run the end-to-end (e2e) tests locally using [kind](https://kind.sigs.k8s.io/), you may need the following additional tools (only tested on Linux; creating a working LoadBalancer on non-Linux environments is TBD, but the [Setting up an Ingress Controller](https://kind.sigs.k8s.io/docs/user/ingress/) guide _may_ work).
+
+1. [Docker Engine](https://docs.docker.com/engine/install/) to run Docker containers
+1. `gcc` -- your distribution should have supplied this
+1. [`kind`](https://kind.sigs.k8s.io/#installation-and-usage) to run Kubernetes clusters locally on Docker
+1. [`kubetest2`](https://github.com/kubernetes-sigs/kubetest2) to set up Kubernetes clusters from the test scripts
+1. [`kntest`](https://github.com/knative/test-infra/tree/main/kntest) for additional utilities wrapping `kubetest2`
+1. [`gcloud`](https://cloud.google.com/sdk/docs/install) and an active registry, for pushing the test images to Google Cloud Registry.
 
 ### Environment setup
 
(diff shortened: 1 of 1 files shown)


nak3:
/cc @carlisia @dprotaso 

Would you mind reviewing? :pray:

DEVELOPMENT.md
```diff
@@ -88,6 +96,32 @@ registry which if you are developing locally you could use `export KO_DOCKER_REP
...
+
+**NOTE** The `dispatch/path`, `host-rewrite`, and `upgrade` tests seem to fail (as of Jan 28) when run locally on Kind. You'll want to add these to the `--skip-tests` flag in `hack/e2e-tests.sh`.
+
+Run the tests from the root directory; the script will create a _new_ Kind cluster and build and install `net-contour` on it, along with building and pushing test images.
+
+```bash
+./test/e2e-tests.sh --cloud-provider kind
+```
+
+**WHILE THE TESTS ARE RUNNING**, install MetalLB in the cluster:
```
dprotaso, line 109: This seems a bit janky and I'm curious why we wouldn't just suggest to spin up a kind/metallb cluster prior to running the tests. 

These scripts support skipping the cluster creation as follows
```
./test/e2e-tests.sh --run-tests
```

dprotaso (commented):
just some minor things

DEVELOPMENT.md
```diff
@@ -88,6 +96,32 @@ registry which if you are developing locally you could use `export KO_DOCKER_REP
 to use the local one on kind. Please see official
 [KO documentation](https://github.com/google/ko#local-publishing-options) for more information.
 
+#### Running end-to-end tests locally
```
carlisia, line 99: First of all, thank you SO much for adding instructions for how to run the tests!


Reading this PR reminded me that this is a file that can use some fixing. Example, that requirement section could go away. For now, my suggestion is that if we group all the testing info together it would make the doc better and it would make the instructions you are adding easier to follow (imo). So, see what happens if you ove a few things around.


First, a note: this PR indicates that there are 2 ways to run the e2e tests, locally vs manually. Once one wraps her head around it, it makes sense. But you will note that I changed that.


Reasoning: because this is a development setup, all tests are to be trigered locally. We don't need to mention that. Also. the "locally" in this PR seems to be the equivalent of 'using scripts to provision a cluster and run all the things", while "manually" seems to mean no setup and scripts needed when there is already a cluster.


So, I changed it to: s/locally/with Kind and s/manually/against a provider.


Maybe it should be s/against a provider/against an existing cluster?


My change iin the naming of those sections might not be super accurate, but I think it's closer in meaning. If it's totally off, feel free to ignore. Or adapt.


1) move the section `### Local Development` to the very end of this file.


2) move the section you added, `#### Running end-to-end tests locally` to under that.


3) move the other section you added, `### Running the e2e tests manually`, to under those two, making it the last section (for now).


Now, let's fix those sections (feel free to use better words, I'm mainly focusing on re-arranging the sections):


- rename `#### Running end-to-end tests locally` to: ### e2e testing`


- add a section: #### Prerequisites
  - move all of the requirements you added to this section and/or subsections below, as appropriate


- add a section: #### Run tests against a provider
  - move the text from your added section `### Running the e2e tests manually` here


- add a section: #### Run tests with Kind
  - move the text from your added section `#### Running end-to-end tests locally` here and/or subsections below, as appropriate


1) Linux


2) MacOS

DEVELOPMENT.md
carlisia, line 109: It seems to be **very important** (lol) to do this while the tests are running. I would move the last paragraph to immediately after this.


-> this paragraph:
> You should do this once the cluster is created, while the tests say "`Waiting until service envoy in namespace contour-external has an external address (IP/hostname)...`".

carlisia (commented):
Woot! Thank you! :panda_face:

DEVELOPMENT.md
evankanderson: I think the double-invocation nature of the script made me unsure how to get a Kind configuration that matched the script setup. I could say "just install Kind and then MetalLB if that works".

evankanderson: I needed to do this while the cluster was running because the script creates the cluster and then waits for all the things (including the LoadBalancer Service) to be ready, but doesn't have a way to run an extra script for setting up the load balancer.

evankanderson (commented):
I updated some of this, but I'm wondering whether we should take out the "set up kind" mode on the e2e test scripts and just say "install Kind with MetalLB" WDYT?

dprotaso:
> but I'm wondering whether we should take out the "set up kind" mode on the e2e test scripts and just say "install Kind with MetalLB" WDYT?

Yeah I'm good with that and we link out to https://kind.sigs.k8s.io/docs/user/loadbalancer

dprotaso:
Unblocking this - any further edits can be done in a follow up

/lgtm
/approve

upodroid (approved):
/easycla

PR #729, opened 2022-02-17
update ytt package name

dprotaso:
/assign @carlisia


carlisia:
Good catch.

/lgtm
/approve

PR #735, opened 2022-03-09
Update Contour to v1.20.1

### LINKED ISSUE #718: Bump Contour version to v1.20.0
[Release Contour v1.20.0 · projectcontour/contour](https://github.com/projectcontour/contour/releases/tag/v1.20.0)

BobyMCbobs:
# Changes

- :broom: update Contour version

/kind enhancement

Fixes #718

**Release Note**

```release-note
Update Contour version to v1.20.1
```


go.mod
```diff
@@ -4,11 +4,11 @@ go 1.16
...
 	github.com/google/go-cmp v0.5.6
-	github.com/projectcontour/contour v1.19.1
+	github.com/projectcontour/contour v1.20.1
 	go.uber.org/zap v1.19.1
-	k8s.io/api v0.22.5
-	k8s.io/apimachinery v0.22.5
-	k8s.io/client-go v0.22.5
+	k8s.io/api v0.23.1
+	k8s.io/apimachinery v0.23.1
+	k8s.io/client-go v0.23.1
```
dprotaso, line 11: Curious if this k8s bump was necessary?

dprotaso: Technically it's fine since our next release (1.4) targets K8s 1.22 (which the k8s 0.23 libs support)

dprotaso:
Looking at the failures we'll need to bump our upstream dependencies first (knative.dev/pkg) 

That's something I can do tomorrow which should unblock this PR.

go.mod
BobyMCbobs: Oh! I didn't mean to do that, I updated the Contour version in `./hack/update-deps.sh` and ran it. Is there any way to not include those changes but to update Contour?

dprotaso: Bumping contour will bump that - since that version becomes the 'min version'.

https://github.com/projectcontour/contour/blob/002720c894dde8f7a397e97ccbd221752f4374de/go.mod#L32-L34

You could use go `replace` directives but it'll probably be a mess in other ways. So I say let's hold this PR and get the K8s bump in first and rebase after that.

dprotaso:
Just a follow up - I tried the bump here: https://github.com/knative/pkg/actions/runs/1964634224

But k8s upstream renamed a struct field which breaks everything - so it'll take longer for that change to trickle across.

I'll ping you on this PR when that's sorted

BobyMCbobs:
> Just a follow up - I tried the bump here: https://github.com/knative/pkg/actions/runs/1964634224

:eyes: 

> But k8s upstream renamed a struct field which breaks everything - so it'll take longer for that change to trickle across.

Dependencies sure can be fun! No worries!

> I'll ping you on this PR when that's sorted

Cheers, @dprotaso!

dprotaso:
Ok - the k8s changes are done and merged at HEAD - you should be able to just update contour and I believe the PR should become much simpler

BobyMCbobs:
> Ok - the k8s changes are done and merged at HEAD - you should be able to just update contour and I believe the PR should become much simpler

Woohoo! Thank you @dprotaso, I've rebased and pushed a commit

dprotaso:
One last thing we need is to bump the go module

https://github.com/knative-sandbox/net-contour/blob/1786a078db3218e92950594a82f0be69b4b10a9c/go.mod#L7

I believe this should work - but it might update other dependencies as well

```bash
$ ./hack/update-dep.sh --upgrade
```

I think it would be simpler to run

```bash
$ go get -u github.com/projectcontour/contour@v1.20.1
$ ./hack/update-deps.sh 
```


the deps script does the go.mod vendoring etc.

BobyMCbobs:
> One last thing we need is to bump the go module
> 
> https://github.com/knative-sandbox/net-contour/blob/1786a078db3218e92950594a82f0be69b4b10a9c/go.mod#L7
> 
> I believe this should work - but it might update other dependencies as well
> 
> ```shell
> $ ./hack/update-dep.sh --upgrade
> ```
> 
> I think it would be simpler to run
> 
> ```shell
> $ go get -u github.com/projectcontour/contour@v1.20.1
> $ ./hack/update-deps.sh 
> ```
> 
> the deps script does the `go mod vendor` and other stuff

@dprotaso, I managed to update things in 3b5293e.
There seems to be a build error if building using Go 1.16
```
🐚 go build -o bin/controller cmd/controller/main.go 
# sigs.k8s.io/json/internal/golang/encoding/json
vendor/sigs.k8s.io/json/internal/golang/encoding/json/encode.go:1249:12: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
vendor/sigs.k8s.io/json/internal/golang/encoding/json/encode.go:1255:18: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
```

which doesn't occur on Go 1.17. Should I also update the required Go version?

dprotaso:
Yup the newer k8s libs require go 1.17

dprotaso:
looks like there's still a diff in your `go.mod` file

https://github.com/knative-sandbox/net-contour/runs/5650184631?check_suite_focus=true

BobyMCbobs:
> looks like there's still a diff in your `go.mod` file
> 
> https://github.com/knative-sandbox/net-contour/runs/5650184631?check_suite_focus=true

@dprotaso, I ran `./hack/update-codegen.sh` and it seems to have updated the go.mod. Hopefully this time!

dprotaso:
/lgtm
/approve

🎉  thanks for the help!

PR #782, opened 2022-07-18
Drop use of the contour operator

dprotaso:
Looks like we weren't using it in our kind e2e tests 

/assign @carlisia


carlisia:
👍 thank you!

/lgtm
/approve

PR #818, opened 2022-10-07
Bump k8s version in kind

nak3:
As per title, this patch follows up https://github.com/knative/pkg/pull/2595.


nak3:
There is a metallb issue. I am not familiar with it and so defer to VMware team.

psschwei (commented):
/lgtm

PR #834, opened 2022-11-24
drop string literal must be 'ALL'

dprotaso:


nak3:
/lgtm

I wondered if `seccompProfile` is needed but as per your comment https://github.com/knative/serving/issues/13471#issuecomment-1326892448 I agree that we should not add it until min k8s version to 1.25.

dprotaso:
/override "e2e tests (v1.23.12, ./test/conformance)"

1.23 isn't supported

PR #844, opened 2022-12-09
Add app.k8s.io labels to the deployment - not just the pod template

### LINKED ISSUE #784: Kubernetes labels not being added to release artifact
For some reason, when generating a release artifact, net-contour is including the k8s labels under `spec.template.metadata.labels`:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: net-contour-controller
  namespace: knative-serving
  labels:
spec:
  replicas: 1
  selector:
    matchLabels:
      app: net-contour-controller
  template:
    metadata:
      labels:
        app: net-contour-controller
        app.kubernetes.io/component: net-contour
        app.kubernetes.io/name: knative-serving
        app.kubernetes.io/version: "1.6.0"
```
Compare that with what net-kourier (corrrectly) generates:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: net-kourier-controller
  namespace: knative-serving
  labels:
    networking.knative.dev/ingress-provider: kourier
    app.kubernetes.io/component: net-kourier
    app.kubernetes.io/version: "1.6.0"
    app.kubernetes.io/name: knative-serving
spec:
  replicas: 1
  selector:
    matchLabels:
      app: net-kourier-controller
  template:
    metadata:
      labels:
        app: net-kourier-controller
```

The fix is probably going to be around this line:
https://github.com/knative-sandbox/net-contour/blob/023772c9c2ed7fae5ac4e47a097a78f76b356744/hack/release.sh#L40

/bug
/good-first-issue

dprotaso:
Fixes #784 

/assign @nak3

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -31,6 +31,9 @@ spec:
     metadata:
       labels:
         app: net-contour-controller
+        app.kubernetes.io/component: net-contour
+        app.kubernetes.io/name: knative-serving
+        app.kubernetes.io/version: devel
     spec:
       serviceAccountName: controller
       containers:
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -19,6 +19,9 @@ metadata:
   namespace: knative-serving
   labels:
     networking.knative.dev/ingress-provider: contour
+    app.kubernetes.io/component: net-contour
+    app.kubernetes.io/name: knative-serving
+    app.kubernetes.io/version: devel
 spec:
   replicas: 1
   selector:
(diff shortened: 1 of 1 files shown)


config/controller.yaml
```diff
@@ -28,9 +31,6 @@ spec:
     metadata:
       labels:
         app: net-contour-controller
-        app.kubernetes.io/component: net-contour
-        app.kubernetes.io/name: knative-serving
```
nak3, line 32: Please allow me to double check. We don't need to keep these labels?
Other components like serving/controller has it https://github.com/knative/serving/blob/main/config/core/deployments/controller.yaml

dprotaso: Good catch - I realize we'd probably want these labels on the Pod as well.

It may be worth creating a kourier issue then

Issue #848, opened 2023-01-05
Too much TLS?


evankanderson:
It appears that net-contour will "promote" internal names like `foo.my-ns.svc` and `foo.my-ns.svc.cluster.local` to be HTTPS (and redirect HTTP callers to HTTPS) when combining the following features:

* `default-tls-secret` is set to anything (causing the `route[*].virtualhost.tls` field to be set)
* `http-protocol` set to `Redirected` or unset default (default is `Redirected`)

This is because [`resources/httpproxy.go`](https://github.com/knative-sandbox/net-contour/blob/main/pkg/reconciler/contour/resources/httpproxy.go) computes a [`permitInsecure` field](https://projectcontour.io/docs/main/config/tls-termination/) and adds some annotations to the internal version of the HTTPProxy, but it doesn't recognize the cluster-internal URLs as needing any special TLS handling, which they probably do.

/good-first-issue

This should be relatively easy to add a test for, I don't know whether we'd want / need to add a flag to enable the old (broken) behavior; my guess is "no".

PR #856, opened 2023-01-20
When ingress visibility is ClusterLocal then permitInsecure

jsanin-vmw:
# Changes

- When ingress visibility is ClusterLocal then permitInsecure
- Unit testing 

/kind bug

Fixes https://github.com/knative-sandbox/net-contour/issues/848

```release-note
Fix cluster local traffic when we set a default certificate is set 
```


dprotaso:
/lgtm
/approve

:tada:

thanks @jsanin-vmw

PR #860, opened 2023-01-31
Internal Encryption fixes

### LINKED ISSUE #862: Enabling Internal Encryption breaks DomainMappings when using Contour
## What version of Knative?

Relocating from https://github.com/knative/serving/issues/13659 since this is just a Contour issue.

Working off main branch of knative-serving, net-contour

Discovered while trying to add internal encryption e2e tests for net-contour here: https://github.com/knative/serving/pull/13536

## Expected Behavior

When I create a DomainMapping for my Knative Service when Internal Encryption is enabled, I am able to reach the KService successfully.

## Actual Behavior

DomainMappings fail to become ready, get stuck in "EndpointsNotReady"

net-contour controller says:
```
{"severity":"ERROR","timestamp":"2023-01-26T21:30:40.304820765Z","logger":"net-contour-controller","caller":"status/status.go:404","message":"Probing of http://hello.gen-14.hello.clay.tanzu.biz.default.net-contour.invalid failed, IP: 10.24.2.35:8080, ready: false, error: unexpected status code: want 200, got 503 (depth: 0)","commit":"e458d29-dirty","knative.dev/controller":"knative.dev.net-contour.pkg.reconciler.contour.Reconciler","knative.dev/kind":"networking.internal.knative.dev.Ingress","knative.dev/traceid":"724fb06d-90db-49cf-917e-a664dd798cb8","knative.dev/key":"default/hello.clay.tanzu.biz--ep","stacktrace":"knative.dev/networking/pkg/status.(*Prober).processWorkItem\n\tknative.dev/networking@v0.0.0-20221202133217-891aac251fc2/pkg/status/status.go:404\nknative.dev/networking/pkg/status.(*Prober).Start.func1\n\tknative.dev/networking@v0.0.0-20221202133217-891aac251fc2/pkg/status/status.go:289"}
```

I also see this in envoy logs:
```
[2023-01-26 21:34:40.776][19][debug][router] [source/common/router/router.cc:1212] [C49017][S13478037937466377759] upstream reset: reset reason: connection failure, transport failure reason: TLS error: 268435703:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
[2023-01-26 21:34:40.776][19][debug][http] [source/common/http/filter_manager.cc:905] [C49017][S13478037937466377759] Sending local reply with details upstream_reset_before_response_started{connection_failure,TLS_error:_268435703:SSL_routines:OPENSSL_internal:WRONG_VERSION_NUMBER}
[2023-01-26 21:34:40.776][19][debug][http] [source/common/http/conn_manager_impl.cc:1551] [C49017][S13478037937466377759] encoding headers via codec (end_stream=false):
':status', '503'
'content-type', 'text/plain'
'content-encoding', 'gzip'
'vary', 'Accept-Encoding'
'date', 'Thu, 26 Jan 2023 21:34:40 GMT'
'server', 'envoy'
```

```
[2023-01-26 21:34:40.735][21][debug][conn_handler] [source/server/active_tcp_listener.cc:147] [C49205] new connection from 10.24.2.43:43224
[2023-01-26 21:34:40.735][21][debug][http] [source/common/http/conn_manager_impl.cc:306] [C49205] new stream
[2023-01-26 21:34:40.735][21][debug][http] [source/common/http/conn_manager_impl.cc:930] [C49205][S163296292697216300] request headers complete (end_stream=true):
':authority', 'hello.gen-3.hello.claysreallyverylongtestineee50218ef4390e47e8e913ebbbebaf8.default.net-contour.invalid'
':path', '/healthz'
':method', 'GET'
'user-agent', 'Knative-Ingress-Probe'
'k-network-hash', 'override'
'k-network-probe', 'probe'
'accept-encoding', 'gzip'
...

[2023-01-26 21:34:40.735][21][debug][http] [source/common/http/conn_manager_impl.cc:913] [C49205][S163296292697216300] request end stream
[2023-01-26 21:34:40.735][21][debug][connection] [./source/common/network/connection_impl.h:92] [C49205] current connecting state: false
[2023-01-26 21:34:40.735][21][debug][router] [source/common/router/router.cc:470] [C49205][S163296292697216300] cluster 'default/hello/80/a67dfba3e6' match for URL '/healthz'
[2023-01-26 21:34:40.735][21][debug][router] [source/common/router/router.cc:678] [C49205][S163296292697216300] router decoding headers:
':authority', 'hello.default.svc.cluster.local'
':path', '/healthz'
':method', 'GET'
':scheme', 'http'
'user-agent', 'Knative-Ingress-Probe'
'k-network-probe', 'probe'
'accept-encoding', 'gzip'
'x-forwarded-for', '10.24.2.43'
'x-forwarded-proto', 'http'
'x-envoy-internal', 'true'
'x-request-id', '9d510376-f3c0-4a55-962d-a1f5a9f0ebe4'
'k-network-hash', 'dc12e833d98a355da2775ad80b3ae02658ed076ec3da0d5670b05f377f36f39e'
'x-request-start', 't=1674768880.735'
...
[2023-01-26 21:34:40.736][21][debug][router] [source/common/router/router.cc:1212] [C49205][S163296292697216300] upstream reset: reset reason: connection failure, transport failure reason: TLS error: 268435703:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
[2023-01-26 21:34:40.736][21][debug][pool] [source/common/conn_pool/conn_pool_base.cc:453] invoking idle callbacks - is_draining_for_deletion_=false
[2023-01-26 21:34:40.758][21][debug][router] [source/common/router/router.cc:1796] [C49205][S163296292697216300] performing retry
...
[2023-01-26 21:34:40.759][21][debug][router] [source/common/router/router.cc:1212] [C49205][S163296292697216300] upstream reset: reset reason: connection failure, transport failure reason: TLS error: 268435703:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
...
[2023-01-26 21:34:40.796][21][debug][router] [source/common/router/router.cc:1796] [C49205][S163296292697216300] performing retry
...
[2023-01-26 21:34:40.798][21][debug][router] [source/common/router/router.cc:1212] [C49205][S163296292697216300] upstream reset: reset reason: connection failure, transport failure reason: TLS error: 268435703:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
...
[2023-01-26 21:34:40.798][21][debug][http] [source/common/http/filter_manager.cc:905] [C49205][S163296292697216300] Sending local reply with details upstream_reset_before_response_started{connection_failure,TLS_error:_268435703:SSL_routines:OPENSSL_internal:WRONG_VERSION_NUMBER}
[2023-01-26 21:34:40.798][21][debug][http] [source/common/http/conn_manager_impl.cc:1551] [C49205][S163296292697216300] encoding headers via codec (end_stream=false):
':status', '503'
'content-type', 'text/plain'
'content-encoding', 'gzip'
'vary', 'Accept-Encoding'
'date', 'Thu, 26 Jan 2023 21:34:40 GMT'
'server', 'envoy'
```

When I try this out with AutoTLS enabled, the domainmappings become ready, but I still get the error when I try hitting the endpoint.
```
upstream connect error or disconnect/reset before headers. retried and the latest reset reason: connection failure, transport failure reason: TLS error: 268435703:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
```

## Steps to Reproduce the Problem

1. enable internal encryption in config-network
2. deploy a simple hello world knative service
3. set up a clusterdomainclaim for your new domain
4. create a domainmapping

## Analysis

I think the problem is in part due to fact that DomainMappings point you back at the envoy.
If you look at the DAG, you can see all the routes point to a service on port 443. However, the one for hello goes to 80:

(DAG output comes from the Contour controller, [see here](https://projectcontour.io/docs/v1.22.1/troubleshooting/contour-graph/))
contour-dag-encryption

That service spec looks like this:
```
apiVersion: v1
kind: Service
metadata:
  ...
  name: hello
  namespace: default
spec:
  clusterIP: None
  clusterIPs:
  - None
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  - IPv6
  ipFamilyPolicy: RequireDualStack
  ports:
  - name: http2
    port: 80
    protocol: TCP
    targetPort: 80
  sessionAffinity: None
  type: ClusterIP
```

Internal encryption was implemented so that ports named http2 are h2c when internal encryption is disabled, h2 when enabled.

This means that the HTTPProxy defines hitting the hello service on port 80 with h2 protocol.

So when Envoy tries to make the call, it uses https (for h2), but hits the http listener on Envoy.

When you put autotls on, there is at least a listener for 443 now, but it doesn't have the route data to deal with the request (since svc.cluster.local domains don't get TLS).

## Suggestion

I think one way around this is to use the internal encryption secrets for the ClusterLocal visibility domains when internal encryption is enabled. That way you get a listener on 443 for those domains. Then you'd need to change the svc to also use 443. The trouble with that is that it is kinda venturing towards TLS for ClusterLocal routes, which is probably a big undertaking.

I suppose another, more simple option is to make the calls from the envoy back to itself not use encryption. But to me, that seems like leaving a hole in the internal encryption path.

KauzClay:
# Changes

:bug: Fixes #862
- Unencrypt request back to envoy for domainmappings when internal encryption is enabled
- net-contour currently sets the protocol as h2 for services with ports named http2 when internal encryption. This change will set the protocol as h2c if the kingress has domainmapping labels.
- This isn't the perfect fix, but it does at least make domainmappings work properly when using internal encryption. I hope this is enough for the beta stage of these features.

:bug: internal encryption breaks http01 challenges
- Don't encrypt traffic to http challenge endpoint, use http instead
- The current implementation tries to use tls to connect to the http01 challenge endpoint when internal encryption is enabled, auto-tls is enabled, and httpOption is redirected, 
- This path is currently hidden by flawed cert-manager behavior that is being addressed [here](https://github.com/knative-sandbox/net-certmanager/pull/481). So long as this fix is included before or in the same release as the cert-manager changes, I don't think anyone will run into this.

 

**Release Note**

```release-note
Fixes issue where DomainMappings do not become ready when Internal Encryption is enabled.
```

**Docs**

```docs
N/A
```

### DIFF
--- a/pkg/reconciler/contour/contour_test.go
+++ b/pkg/reconciler/contour/contour_test.go
@@ -542,6 +542,37 @@ func TestReconcileInternalEncryption(t *testing.T) {
 			},
 			Name: "name--ep",
 		}},
+	}, {
+		Name: "first reconcile domainmapping ingress (endpoints probe succeeded)",
+		Key:  "ns/dm-name",
+		Objects: append([]runtime.Object{
+			ing("dm-name", "ns", withDomainMappingSpec, withContour),
+			mustMakeProbe(t, ing("dm-name", "ns", withDomainMappingSpec, withContour), makeItReady),
+		}, servicesAndEndpoints...),
+		WantCreates: mustMakeProxiesWithConfig(t, ing("dm-name", "ns", withDomainMappingSpec, withContour), internalEncryptionConfig),
+		WantStatusUpdates: []clientgotesting.UpdateActionImpl{{
+			Object: ing("dm-name", "ns", withDomainMappingSpec, withContour, func(i *v1alpha1.Ingress) {
+				// These are the things we expect to change in status.
+				i.Status.InitializeConditions()
+				i.Status.MarkNetworkConfigured()
+				i.Status.MarkLoadBalancerReady(
+					[]v1alpha1.LoadBalancerIngressStatus{{
+						DomainInternal: publicSvc,
+						IP:             publicSvcIP,
+					}},
+					[]v1alpha1.LoadBalancerIngressStatus{{
+						DomainInternal: privateSvc,
+						IP:             privateSvcIP,
+					}})
+			}),
+		}},
+		WantDeletes: []clientgotesting.DeleteActionImpl{{
+			ActionImpl: clientgotesting.ActionImpl{
+				Namespace: "ns",
+				Resource:  v1alpha1.SchemeGroupVersion.WithResource("ingresses"),
+			},
+			Name: "dm-name--ep",
+		}},
 	}}
 
 	table.Test(t, MakeFactory(func(ctx context.Context, listers *Listers, cmw configmap.Watcher) controller.Reconciler {
@@ -979,6 +1010,32 @@ func withTLSServiceSpec(i *v1alpha1.Ingress) {
 							ServiceName:      tlsServiceName,
 							ServiceNamespace: i.Namespace,
 							ServicePort:      intstr.FromInt(443),
+						},
+						Percent: 100,
+					}},
+				}},
+			},
+		}},
+	}
+}
+
+func withDomainMappingSpec(i *v1alpha1.Ingress) {
+	i.Spec = v1alpha1.IngressSpec{
+		HTTPOption: v1alpha1.HTTPOptionEnabled,
+		Rules: []v1alpha1.IngressRule{{
+			Hosts:      []string{"dm.example.com"},
+			Visibility: v1alpha1.IngressVisibilityExternalIP,
+			HTTP: &v1alpha1.HTTPIngressRuleValue{
+				Paths: []v1alpha1.HTTPIngressPath{{
+					RewriteHost: "name.ns.svc.cluster.local",
+					Splits: []v1alpha1.IngressBackendSplit{{
+						AppendHeaders: map[string]string{
+							"K-Original-Host": "dm.example.com",
+						},
+						IngressBackend: v1alpha1.IngressBackend{
+							ServiceName:      "doo",
+							ServiceNamespace: i.Namespace,
+							ServicePort:      intstr.FromInt(80),
 						},
 						Percent: 100,
 					}},
--- a/pkg/reconciler/contour/resources/constants.go
+++ b/pkg/reconciler/contour/resources/constants.go
@@ -45,6 +45,10 @@ const (
 	//Contour's Envoys to encrypt the traffic to the activator.
 	InternalEncryptionProtocol   = "tls"
 	InternalEncryptionH2Protocol = "h2"
+
+	//HttpChallengePath is the path that gets added to routes when using
+	//auto-TLS with an http01 solver as the issuer.
+	HTTPChallengePath = "/.well-known/acme-challenge"
 )
 
 // These are the annotations which are optionally set in ksvc/ingress
--- a/pkg/reconciler/contour/resources/httpproxy.go
+++ b/pkg/reconciler/contour/resources/httpproxy.go
@@ -32,6 +32,7 @@ import (
 	"knative.dev/net-contour/pkg/reconciler/contour/config"
 	"knative.dev/networking/pkg/apis/networking/v1alpha1"
 	netcfg "knative.dev/networking/pkg/config"
+	netheader "knative.dev/networking/pkg/http/header"
 	"knative.dev/networking/pkg/ingress"
 	"knative.dev/pkg/kmeta"
 	"knative.dev/pkg/network"
@@ -170,11 +171,16 @@ func MakeHTTPProxies(ctx context.Context, ing *v1alpha1.Ingress, serviceToProtoc
 				postSplitHeaders := &v1.HeadersPolicy{
 					Set: make([]v1.HeaderValue, 0, len(split.AppendHeaders)),
 				}
+
+				hasOriginalHostKey := false
 				for key, value := range split.AppendHeaders {
 					postSplitHeaders.Set = append(postSplitHeaders.Set, v1.HeaderValue{
 						Name:  key,
 						Value: value,
 					})
+					if key == netheader.OriginalHostKey {
+						hasOriginalHostKey = true
+					}
 				}
 				if len(postSplitHeaders.Set) > 0 {
 					sort.Slice(postSplitHeaders.Set, func(i, j int) bool {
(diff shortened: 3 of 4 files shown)


pkg/reconciler/contour/contour.go
```diff
@@ -332,3 +333,11 @@ func (r *Reconciler) lbStatus(ctx context.Context, vis v1alpha1.IngressVisibilit
 	}
 	return
 }
+
+func isDomainMapping(ing *v1alpha1.Ingress) bool {
+	_, hasDmUIDLabelKey := ing.Labels[serving.DomainMappingUIDLabelKey]
```
KauzClay, line 338: I had to bring in a fair amount of serving code just to use these labels. Maybe there is a better way to import it that doesn't bring in so much?

evankanderson: What you probably want to do is to either copy the constants here, or move them into `networking`.  When you have `net-contour` import `serving` directly, it makes it hard to cut a release, because the `net-contour` package can't be cut until the `serving` package is tagged.

Alternatively, you could look at the `OwnerReferences` field to see whether the owner is a Knative Route or a DomainMapping.

The third option would be to add some explicit field to the KIngress indicating which type it is.

The fourth option would be to look at the `spec.rules` -- if there's a single `rule` with `visibility: externalIP`, that must be a DomainMapping, because a Route always has an internal mapping: https://github.com/knative/serving/blob/main/pkg/reconciler/route/resources/ingress.go#L141

KauzClay: ah okay, thanks for explaining that about importing serving. 

I didn't even think about the owner reference, I think I will go for that. It seems the least intrusive but still pretty explicit. Then I can avoid having duplicated constants or adding more fields to the kingress.

dprotaso: DomainMapping just sets `HostRewrite` on the KIngress - https://github.com/knative/networking/blob/2473e65d69206ed544f32e47f994cf3635bebdbe/pkg/apis/networking/v1alpha1/ingress_types.go#L208-L212

You could look at that spec property and the target k8s service (it pointing to a contour's envoy service)

KauzClay: ah, well maybe not, the type definition for domainmappings is in the serving repo.

evankanderson:
> I wonder if TLS passthrough makes more sense.

I like that, except that we have no KIngress construct to describe this (and it's not clear that we should).

pkg/reconciler/contour/contour.go
evankanderson: I think you can define a constant for DomainMapping here, or add a helper in `networking` to determine if a KIngress represents a DomainMapping.

It might be more robust, though, to look for the `rewriteHeader` field and the presence of `knative.dev/networking/pkg/http/header.OriginalHostKey` as a field in `AppendHeaders`.

KauzClay:
Hey @evankanderson @dprotaso , do you have any other thoughts on this one? 

To me it sounds like any actions regarding Dave's comment
>This PR makes me realize that we need a special consideration for traffic going through external envoy => internal envoy and ensuring that hop is encrypted - right now we're setting it to h2c.

probably should happen in another PR. IS that what you had in mind?

dprotaso:
> probably should happen in another PR. Is that what you had in mind?

Yeah let's make an new issue and do it as a follow up

/lgtm
/approve

PR #861, opened 2023-02-01
bump contour to 1.24.0

KauzClay:
# Changes

- :gift: Bring in Contour [1.24.0](https://github.com/projectcontour/contour/releases/tag/v1.24.0)

/kind enhancement

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


dprotaso:
Can we bump the version in the `go.mod` as well?

if you run `./hack/update-deps.sh --upgrade` that should pull in latest contour changes

KauzClay:
> if you run ./hack/update-deps.sh --upgrade that should pull in latest contour changes

Okay I tried that out. A whole lot changed, is that expected?

KauzClay:
I looked into the error in the build task:
```
Error: vendor/knative.dev/pkg/client/injection/kube/client/client.go:215:30: cannot use (*wrapClient)(nil) (value of type *wrapClient) as type kubernetes.Interface in variable declaration:
	*wrapClient does not implement kubernetes.Interface (missing AdmissionregistrationV1alpha1 method)
```

It looks like this bump is pulling in `k8s.io/client-go v0.26.1`. If you search for admissionregistration in the [k8s 1.26 release changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.26.md#api-change) it looks like something was added.

It seems like the knative code affected is generated code. Does some sort of update to knative.dev/pkg for k8s 1.26 have to happen first?

dprotaso:
> It seems like the knative code affected is generated code. Does some sort of update to knative.dev/pkg for k8s 1.26 have to happen first?

Yeah though unfortunately we want to keep our libs at 0.25 since we're still targeting K8s v1.24.

You'll want to add `replace` directives to the `go.mod` pinning the k8s dependencies to 0.25.x

go.mod
```diff
@@ -2,96 +2,102 @@ module knative.dev/net-contour
...
+	google.golang.org/genproto v0.0.0-20230117162540-28d6b9783ac4 // indirect
+	google.golang.org/grpc v1.52.3 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	k8s.io/apiextensions-apiserver v0.25.4 // indirect
-	k8s.io/code-generator v0.25.4 // indirect
+	k8s.io/apiextensions-apiserver v0.26.1 // indirect
+	k8s.io/code-generator v0.26.1 // indirect
```
dprotaso, line 96: you'll want to pin these indirect dependencies as well

PR #878, opened 2023-02-23
Fix DomainMappings when internal-encryption is enabled

### LINKED ISSUE #862: Enabling Internal Encryption breaks DomainMappings when using Contour

KauzClay:
# Changes

- :bug: Fix bug
- Set proto to h2c when ports named http2 use port 80, even when internal encryption is on. 
- the previous fix in #860 expected the K-Original-Host header to be present in the appendheaders section. However, it turns out that isn't added to the endpoint probe. This means that the ep would use h2 on port 80 when it was for DomainMappings.
- Instead of adding the k-original-host header to the endpoint probes, or only checking for the rewriteHost field, I think the solution I'm proposing in the PR is simpler, and gets at the root of the problem: traffic to port 80 shouldn't try to be encrypted.
- See this discussion for more context: https://github.com/knative-sandbox/net-contour/pull/860#discussion_r1099192228

/kind bug

Fixes #862 

**Release Note**

```release-note
Fixes issue where DomainMappings do not become ready when Internal Encryption is enabled.
```

**Docs**

```docs
N/A
```

### DIFF
--- a/pkg/reconciler/contour/contour.go
+++ b/pkg/reconciler/contour/contour.go
@@ -189,7 +189,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) r
 		for _, port := range svc.Spec.Ports {
 
 			if port.Name == networking.ServicePortNameH2C {
-				if cfg.Network != nil && cfg.Network.InternalEncryption {
+				if cfg.Network != nil && cfg.Network.InternalEncryption && port.Port != networking.ServiceHTTPPort {
 					serviceToProtocol[name] = resources.InternalEncryptionH2Protocol
 					logger.Debugf("marked an http2 svc %s as h2 for internal encryption", name)
 				} else {
--- a/pkg/reconciler/contour/contour_test.go
+++ b/pkg/reconciler/contour/contour_test.go
@@ -548,7 +548,7 @@ func TestReconcileInternalEncryption(t *testing.T) {
 		Objects: append([]runtime.Object{
 			ing("dm-name", "ns", withDomainMappingSpec, withContour),
 			mustMakeProbe(t, ing("dm-name", "ns", withDomainMappingSpec, withContour), makeItReady),
-		}, servicesAndEndpoints...),
+		}, tlsServiceAndEndpoint...),
 		WantCreates: mustMakeProxiesWithConfig(t, ing("dm-name", "ns", withDomainMappingSpec, withContour), internalEncryptionConfig),
 		WantStatusUpdates: []clientgotesting.UpdateActionImpl{{
 			Object: ing("dm-name", "ns", withDomainMappingSpec, withContour, func(i *v1alpha1.Ingress) {
@@ -800,7 +800,7 @@ var (
 		&corev1.Service{
 			ObjectMeta: metav1.ObjectMeta{
 				Namespace: "ns",
-				Name:      tlsServiceName,
+				Name:      tlsRevisionServiceName,
 			},
 			Spec: corev1.ServiceSpec{
 				Ports: []corev1.ServicePort{{
@@ -808,6 +808,18 @@ var (
 					Port:       443,
 					TargetPort: intstr.FromInt(8443),
 					Protocol:   corev1.ProtocolTCP,
+				}},
+			},
+		},
+		&corev1.Service{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: "ns",
+				Name:      tlsServiceName,
+			},
+			Spec: corev1.ServiceSpec{
+				Ports: []corev1.ServicePort{{
+					Name: "http2",
+					Port: 80,
 				}},
 			},
 		},
@@ -835,7 +847,7 @@ var (
 		&corev1.Endpoints{
 			ObjectMeta: metav1.ObjectMeta{
 				Namespace: "ns",
-				Name:      tlsServiceName,
+				Name:      tlsRevisionServiceName,
 			},
 			Subsets: []corev1.EndpointSubset{{
 				Addresses: []corev1.EndpointAddress{{
@@ -847,14 +859,32 @@ var (
 				}},
 			}},
 		},
+		&corev1.Endpoints{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: "ns",
+				Name:      tlsServiceName,
+			},
+			Subsets: []corev1.EndpointSubset{{
+				Addresses: []corev1.EndpointAddress{{
+					IP: "192.168.1.1",
+				}},
+				Ports: []corev1.EndpointPort{{
+					Name:     "http2",
+					Port:     80,
+					Protocol: corev1.ProtocolTCP,
+				}},
+			}},
+		},
 	}
 	tlsServiceAndEndpoint = append(append([]runtime.Object{}, tlsService...), tlsEndpoint...)
 
-	h2cServiceName    = "doo"
-	tlsServiceName    = "tlsService"
-	serviceToProtocol = map[string]string{
-		h2cServiceName: "h2c",
-		tlsServiceName: resources.InternalEncryptionProtocol,
+	h2cServiceName         = "doo"
+	tlsServiceName         = "tlsService"
+	tlsRevisionServiceName = tlsServiceName + "-00001"
+	serviceToProtocol      = map[string]string{
+		h2cServiceName:         "h2c",
+		tlsServiceName:         "h2c",
+		tlsRevisionServiceName: resources.InternalEncryptionProtocol,
 	}
 )
 
@@ -1007,7 +1037,7 @@ func withTLSServiceSpec(i *v1alpha1.Ingress) {
 				Paths: []v1alpha1.HTTPIngressPath{{
 					Splits: []v1alpha1.IngressBackendSplit{{
 						IngressBackend: v1alpha1.IngressBackend{
-							ServiceName:      tlsServiceName,
+							ServiceName:      tlsRevisionServiceName,
 							ServiceNamespace: i.Namespace,
 							ServicePort:      intstr.FromInt(443),
 						},
@@ -1033,7 +1063,7 @@ func withDomainMappingSpec(i *v1alpha1.Ingress) {
 							"K-Original-Host": "dm.example.com",
 						},
 						IngressBackend: v1alpha1.IngressBackend{
-							ServiceName:      "doo",
+							ServiceName:      tlsServiceName,
 							ServiceNamespace: i.Namespace,
 							ServicePort:      intstr.FromInt(80),
 						},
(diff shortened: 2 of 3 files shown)


dprotaso:
> traffic to port 80 shouldn't try to be encrypted.

Right now the queue proxy will spin up two http servers (TLS & non-TLS) - but long term I don't think we should continue to spin up the non-TLS server when internal encryption is on.

So I wonder if the _right_ thing to do here is to make sure we start probing the TLS ports as a solution vs. continuing to probe on an unencrypted port 80

KauzClay:
> > traffic to port 80 shouldn't try to be encrypted.
> 
> Right now the queue proxy will spin up two http servers (TLS & non-TLS) - but long term I don't think we should continue to spin up the non-TLS server when internal encryption is on.
> 
> So I wonder if the _right_ thing to do here is to make sure we start probing the TLS ports as a solution vs. continuing to probe on an unencrypted port 80

@dprotaso Maybe I misunderstood when I was poking at it, but the trouble I've run into is that when using a domain mapping, the service that gets probed is that top level service that points to the envoy. So if we update that service to have a 443 entry, then we also need the envoy to listen on 443. And then that starts creeping into the territory of adding TLS for internal routes.

My cluster is all messed up right now, but once I fix it I'll try to get a better example of what I'm trying to say

dprotaso:
> The service that gets probed is that top level service that points to the envoy.
> ...
> So if we update that service to have a 443 entry, then we also need the envoy to listen on 443. And then that starts creeping into the territory of adding TLS for internal routes.

When internal TLS is turned on everything goes through the activator so I believe that would be the target of the probing. I believe TLS for internal routes should 'just work'

> And my goal for this PR was just to get the current broken implementation into a workable state.

We do want all the hops encrypted though so that would included traffic from external proxy => internal proxy. I'm flexble on the path we take to get there so let me know what you think.

dprotaso:
Sorry the target of the probing = what responds to the probe

KauzClay:
kinda left this one hanging, but, I think the work to add https to clusterLocal routes here (https://github.com/knative-sandbox/net-certmanager/pull/538) might help in this scenario. 

As that moves forward, I'll see if I can adjust this PR accordingly

PR #912, opened 2023-05-22
refactor: use new dataplane-trust flag and new certificate name ServingRoutingCertName

KauzClay:
# Changes

- :broom: following on from https://github.com/knative/networking/pull/778
- internal-encryption flag is deprecated, remove usage. Replace with dataplane-trust flag.
- dataplane-trust = minimal is equivalent to what is implemented for internal-encryption = enabled
- Also use the new secret name that the certs will be stored in, [ServingRoutingCertName](https://github.com/knative/networking/blob/0dbe4f9cade877c06f7af2acccaa9e978c7dd5e4/pkg/config/config.go#L78)

/kind cleanup 
/kind deprecation

Fixes #

**Release Note**

```release-note
net-contour respects the dataplane-trust config to enable internal encryption. No longer supports deprecated "internal-encryption" flag
```

**Docs**

```docs
N/A
```

### DIFF
--- a/pkg/reconciler/contour/contour.go
+++ b/pkg/reconciler/contour/contour.go
@@ -40,6 +40,7 @@ import (
 	"knative.dev/net-contour/pkg/reconciler/contour/resources/names"
 	"knative.dev/networking/pkg/apis/networking"
 	"knative.dev/networking/pkg/apis/networking/v1alpha1"
+	netcfg "knative.dev/networking/pkg/config"
 	"knative.dev/networking/pkg/status"
 	"knative.dev/pkg/kmp"
 	"knative.dev/pkg/logging"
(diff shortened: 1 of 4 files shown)


KauzClay:
@davidhadas you may also want to take a look, since you were the one who added these new config options

davidhadas (changes requested):
The PR should consider any trust configuration which is not Disabled as what we used to have as "internal-encryption=enabled". We now have multiple levels  - all of them are to be considered "enabled". In a later PR, we will be more specific to say what are the differences between the levels. 

Please also change the certificate used - envoy should no longer use the deprecated `ServingInternalCertName` and instead use `ServingRoutingCertName`

PR #920, opened 2023-06-02
Bump Contour to v1.25 and bump k8s libs

dprotaso:


dprotaso:
/override "style / suggester / shell"

Made an issue about this - https://github.com/knative/actions/issues/145

PR #923, opened 2023-06-06
Get certificate reconciler from networking instead of control-protocol

ReToCode:
Follow-up for https://github.com/knative/networking/pull/802

## Proposed Changes
* Get certificate reconciler from `networking` instead of `control-protocol`
* Bump all dependencies with `hack/update-deps.sh --upgrade`
* Drop `control-protocol` as a dependency

**Release Note**
NONE


ReToCode:
cc @dprotaso, @KauzClay

PR #974, opened 2023-10-04
Bump k8s api to `v0.27.6` and contour to `1.26.0`

### LINKED ISSUE #960: Bump Contour to v1.26
https://github.com/projectcontour/contour/releases

andrew-su:
# Changes
- Bump contour to `1.26.0`
- Bump k8s API to `0.27.6`
- Force `kube-openapi` to be the same version used by `k8s.io/client-go@0.27.6`.
  - Letting this update to latest (`>= 0.28.0`) causes a dependency issue as `github.com/google/gnostic` had been split out into `github.com/google/gnostic-models`
- Force `code-generator` to be `v0.27.6`.
  - The newer version of `k8s.io/code-generator` executes `generate-internal-groups.sh` which we currently do not set the exec flag from the `knative.dev/hack` module. (To be removed when https://github.com/knative/hack/issues/335 or https://github.com/knative/hack/issues/336 is done)

/kind enhancement

Fixes #960


dprotaso:
@andrew-su all the stuff has merged - so you're good

andrew-su:
Updated.
I had chosen to use the same versions as what we currently have for the other k8s dependencies, however it looks like the codegen still work with:
```
	k8s.io/code-generator => k8s.io/code-generator v0.27.6
	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20230501164219-8b0f38b5fd1f
```

Should I bump it to ^ or leave it for consistency?

dprotaso:
We can match serving with v0.27x deps. Next release we are targeting k8s min version of 1.26 which means we can use 1.27 libs

https://github.com/knative/serving/blob/6703812f131472daa5b7f17e60dd7f49a06c1ead/go.mod#L28

Issue #961, opened 2023-09-15
Bump Go and Kind Versions


dprotaso:
Kind version needs to test version v1.26 to v.1.28
Go version needs to be bumped to v1.21

Similar to: https://github.com/knative/serving/issues/14381

ReToCode:
Closed with https://github.com/knative-extensions/net-contour/pull/977

PR #977, opened 2023-10-11
fix: Bump Go and Kind Versions

MeenuyD:
Kind version change to test version v1.26 to v.1.28
Go version is bumped to v1.21
#961

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ replace (
 require (
 	github.com/google/go-cmp v0.5.9
 	github.com/projectcontour/contour v1.25.0
-	go.uber.org/zap v1.26.0
+	go.uber.org/zap v1.25.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -11,7 +11,7 @@ replace (
 require (
 	github.com/google/go-cmp v0.5.9
 	github.com/projectcontour/contour v1.25.0
-	go.uber.org/zap v1.25.0
+	go.uber.org/zap v1.26.0
```
dprotaso, line 14: same as the other PR - you don't need to bump zap

dprotaso (changes requested):

go.mod
dprotaso: let me reverse this change

PR #1046, opened 2024-02-22
bump-go to v1.22

dprotaso:
Use the reusable action in knative/actions - this will bump go to v1.22

/assign @izabelacg


.github/workflows/kind-e2e.yaml
```diff
@@ -54,18 +50,3 @@ jobs:
         cluster-resources: nodes,namespaces,crds
         namespace-resources: pods,svc,king,httpproxy
         artifact-name: logs-${{ matrix.k8s-version }}
-
-    - name: Post failure notice to Slack
```
dprotaso, line 58: This was something we didn't need

PR #1069, opened 2024-04-03
Add CORS policy to config-contour

izabelacg:
# Changes

- add corsPolicy property to config-contour to allow corsPolicy configuration flow into HTTPProxies


pkg/reconciler/contour/config/contour.go
```diff
@@ -70,6 +83,37 @@ func NewContourFromConfigMap(configMap *corev1.ConfigMap) (*Contour, error) {
...
 	}
 
+	cors, ok := configMap.Data[corsPolicy]
+	if ok {
+		if err := yaml.Unmarshal([]byte(cors), &contourCORSPolicy); err != nil {
+			return nil, err
+		}
+
+		fields := [][]string{
+			contourCORSPolicy.AllowOrigin,
```
dprotaso, line 93: The regexp check isn't applied to `AllowOrigin`

https://github.com/projectcontour/contour/blob/fe259fa4f9060210b4e43e49aab9a5daedf84c94/apis/projectcontour/v1/httpproxy.go#L505-L507

pkg/reconciler/contour/config/contour.go
```diff
@@ -70,6 +83,37 @@ func NewContourFromConfigMap(configMap *corev1.ConfigMap) (*Contour, error) {
...
+			contourCORSPolicy.AllowMethods,
+			contourCORSPolicy.AllowHeaders,
+			contourCORSPolicy.ExposeHeaders,
+		}
+		for _, field := range fields {
+			if len(field) > 0 {
+				var validOption = regexp.MustCompile("^[a-zA-Z0-9!#$%&'*+.^_`|~-]+$")
+				for _, option := range field {
+					if !validOption.MatchString(option) {
+						return nil, fmt.Errorf("option %s is not validly formated", option)
```
dprotaso, line 103: error messages here need to highlight that this is part of the `corsPolicy` attribute

pkg/reconciler/contour/config/contour.go
```diff
@@ -39,6 +40,7 @@ const (
 	defaultTLSSecretConfigKey = "default-tls-secret"
 	timeoutPolicyIdleKey      = "timeout-policy-idle"
 	timeoutPolicyResponseKey  = "timeout-policy-response"
+	corsPolicy                = "corsPolicy"
```
dprotaso, line 43: We use `snake-case` in config maps

```suggestion
	corsPolicy                = "cors-policy"
```

pkg/reconciler/contour/config/contour.go
```diff
@@ -49,18 +51,29 @@ type Contour struct {
 	DefaultTLSSecret      *types.NamespacedName
 	TimeoutPolicyResponse string
 	TimeoutPolicyIdle     string
+	CORSPolicy            *CORSPolicy
```
dprotaso, line 54: Should just use the contour's cors struct

pkg/reconciler/contour/resources/httpproxy.go
```diff
@@ -300,8 +302,20 @@ func MakeHTTPProxies(ctx context.Context, ing *v1alpha1.Ingress, serviceToProtoc
 				}
 
 				hostProxy.Name = kmeta.ChildName(ing.Name+"-"+class+"-", host)
-				hostProxy.Spec.VirtualHost = &v1.VirtualHost{
-					Fqdn: host,
+
+				if cfg.Contour.CORSPolicy != nil {
+					corsPolicy := &v1.CORSPolicy{}
+					tmp, _ := json.Marshal(cfg.Contour.CORSPolicy)
+					json.Unmarshal(tmp, &corsPolicy)
```
dprotaso, line 309: we shouldn't do this marshalling - just use the `v1.CORSPolicy` once in the config

pkg/reconciler/contour/config/contour.go
```diff
@@ -49,18 +51,29 @@ type Contour struct {
...
+}
+
+type CORSPolicy struct {
```
KauzClay, line 57: question:

what do you think importing Contour's definition of the CORS section of the HTTPProxy?

I think that is what you marshal it into anyway later on in `pkg/reconciler/contour/resources/httpproxy.go` on line 307? 

https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/httpproxy.go#L494

KauzClay: oh, didn't see Dave's comments, he said the same thing

pkg/reconciler/contour/config/contour.go
izabelacg: we just switched to contour's struct

pkg/reconciler/contour/config/contour.go
izabelacg: addressed

pkg/reconciler/contour/config/contour.go
izabelacg: addressed

pkg/reconciler/contour/config/contour.go
izabelacg: 👀

pkg/reconciler/contour/config/contour.go
```diff
@@ -70,6 +75,40 @@ func NewContourFromConfigMap(configMap *corev1.ConfigMap) (*Contour, error) {
 		return nil, err
 	}
 
+	cors, ok := configMap.Data[corsPolicy]
+	if ok {
+		if err := yaml.Unmarshal([]byte(cors), &contourCORSPolicy); err != nil {
+			return nil, err
```
dprotaso, line 81: This needs a better message because you won't know the `cors-policy` field wasn't valid

pkg/reconciler/contour/config/contour.go
```diff
@@ -70,6 +75,40 @@ func NewContourFromConfigMap(configMap *corev1.ConfigMap) (*Contour, error) {
...
+			contourCORSPolicy.AllowMethods,
+			contourCORSPolicy.AllowHeaders,
+			contourCORSPolicy.ExposeHeaders,
+		}
+		for _, field := range fields {
+			if len(field) > 0 {
+				var validOption = regexp.MustCompile("^[a-zA-Z0-9!#$%&'*+.^_`|~-]+$")
+				for _, option := range field {
+					if !validOption.MatchString(string(option)) {
+						return nil, fmt.Errorf("%s.allowMethods, %s.allowHeaders, and %s.exposeHeaders have to be validly formatted. Option %s is invalid", corsPolicy, corsPolicy, corsPolicy, option)
```
dprotaso, line 98: this error isn't really useful to the user - should indicate which field is invalid and at which index

pkg/reconciler/contour/config/contour.go
```diff
@@ -70,6 +75,40 @@ func NewContourFromConfigMap(configMap *corev1.ConfigMap) (*Contour, error) {
...
+					}
+				}
+			}
+		}
+
+		if len(contourCORSPolicy.MaxAge) > 0 {
+			_, err := time.ParseDuration(contourCORSPolicy.MaxAge)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse %s.maxAge: %w", corsPolicy, err)
+			}
```
dprotaso, line 108: Not really validating max age here - in theory we would allow negative numbers here which the contour API does not

https://github.com/projectcontour/contour/blob/fe259fa4f9060210b4e43e49aab9a5daedf84c94/apis/projectcontour/v1/httpproxy.go#L523C27-L526

pkg/reconciler/contour/resources/httpproxy.go
```diff
@@ -300,8 +300,16 @@ func MakeHTTPProxies(ctx context.Context, ing *v1alpha1.Ingress, serviceToProtoc
...
+				if cfg.Contour.CORSPolicy != nil {
+					hostProxy.Spec.VirtualHost = &v1.VirtualHost{
+						Fqdn:       host,
+						CORSPolicy: cfg.Contour.CORSPolicy,
+					}
+				} else {
+					hostProxy.Spec.VirtualHost = &v1.VirtualHost{
+						Fqdn: host,
+					}
 				}
```
dprotaso, line 313: Don't need such a complex if statement

```suggestion
				hostProxy.Spec.VirtualHost = &v1.VirtualHost{
					Fqdn:       host,
				}

				if cfg.Contour.CORSPolicy != nil {
					hostProxy.Spec.VirtualHost.CORSPolicy = cfg.Contour.CORSPolicy
				}
```

pkg/reconciler/contour/config/contour.go
krsna-m: `error unmarshaling JSON: while decoding JSON: json: cannot unmarshal number into Go struct field CORSPolicy.allowCredentials of type bool` What more info you looking for?

pkg/reconciler/contour/config/contour.go
krsna-m: Good catch! used their regex.

pkg/reconciler/contour/config/contour.go
dprotaso: Oh cool the yaml library returns a good error message

dprotaso:
See the check
>  Code Style / style / Golang / Auto-format and Check

dprotaso:
thanks 🎉 
/lgtm
/approve

PR #1088, opened 2024-05-13
Set CORS policy values only for when HTTPProxy associated with external visibility

izabelacg:
# Changes

- :broom: Currently, when setting CORS policy in config-contour, all HTTPProxies are modified to include the CORS settings.
This PR modifies the code to set CORS policy values only for when HTTPProxy is associated with external visibility.
- add test to confirm CORS policy settings are not included in HTTPProxy resources with cluster local visibility.

/kind cleanup


ReToCode (approved):
/lgtm
/approve

PR #1095, opened 2024-05-22
Fix nil pointer dereference error when parsing config map

izabelacg:
# Changes

- :bug: Fix nil pointer dereference error when parsing the cors-policy field config-contour config map

PS: Error found while investigating issues in serving tests:
* https://github.com/knative/serving/pull/15204
* https://github.com/knative/serving/pull/15231

/kind bug

**Release Note**

```release-note
Resolved a nil pointer exception triggered by the cors-policy field being set to an empty value.
```

### DIFF
--- a/pkg/reconciler/contour/config/contour.go
+++ b/pkg/reconciler/contour/config/contour.go
@@ -81,7 +81,7 @@ func NewContourFromConfigMap(configMap *corev1.ConfigMap) (*Contour, error) {
 			return nil, err
 		}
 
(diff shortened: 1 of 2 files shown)


dprotaso:
Was there some malformed config in the serving tests?

izabelacg:
> Was there some malformed config in the serving tests?

in the serving e2e script, I set the cors-policy field to empty to disable the feature, which should result in a parsing error. That's why the subsequent tests fail, since the net-contour pod enters in an error state and doesn't recover. I'm waiting on the test PR to confirm if I'm properly removing the field instead of setting it to empty

PR #1149, opened 2024-12-21
bump contour to 1.30.1

dprotaso:
- adjust linter settings
- bump contour to 1.30.1

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,5 +1,6 @@
 run:
-  timeout: 5m
+  timeout: 10m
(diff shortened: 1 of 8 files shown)


dprotaso:
Looks like there's a regression with contour v1.30.1 and a bunch of our tests

dprotaso:
Asked if there are known issues here - https://kubernetes.slack.com/archives/C8XRH2R4J/p1734754791329519

Unfortunately, don't have time to investigate this

PR #1154, opened 2025-01-10
drop replace directives

dprotaso:


dprotaso:
Will do linting in another pass

/override "style / Golang / Lint"

PR #1171, opened 2025-04-11
drop coverage target

dprotaso:
Dropping to 80 - I believe the project is at 85 now.

/assign @skonto

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -4,7 +4,7 @@ coverage:
   status:
     project:
       default:
-        target: 90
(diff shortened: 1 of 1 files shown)


dprotaso:
/hold - something is up - with coverage i don't think we need to lower it

chat https://cloud-native.slack.com/archives/C04LY4M2G49/p1744335175239699

dprotaso:
replacing this PR with this one
https://github.com/knative-extensions/net-contour/pull/1175

PR #1243, opened 2026-02-12
Drop replace directives for incoming k8s deps bump

dprotaso:


dprotaso:
Unblocks - https://github.com/knative/pkg/pull/3316

PR #1301, opened 2026-08-31
Preserve backend ports and host rewrites in endpoint probes

kahirokunn:
# Problem

Before switching a KIngress to new Envoy programming, net-contour creates a temporary `--ep` KIngress that keeps the current and previous backends available while Envoy loads the new endpoint configuration.

For example, an update can replace this previously programmed HTTPProxy route:

```yaml
routes:
- requestHeadersPolicy:
    set:
    - name: Host
      value: app.default.svc.cluster.local
  services:
  - name: app
    port: 80
```

with this KIngress backend:

```yaml
splits:
- serviceName: app
  servicePort: 8080
```

The endpoint probe grouped backends only by Service name, so `app:8080` replaced `app:80` in the temporary route. When a previous Service was restored, its route-level `Host` rewrite was also lost. The probe could therefore omit the previous Envoy cluster or forward its synthetic `*.net-contour.invalid` host to a backend that requires the internal host, leaving the KIngress in `EndpointsNotReady`.

The Host rewrite case is exercised by [knative/serving#16685](https://github.com/knative/serving/pull/16685), which changes a DomainMapping KIngress from a cluster-local Service backend to direct Revision Service backends.

# Changes

Key endpoint-probe backends by Service name and port, include the port in each synthetic probe host, and restore the previous HTTPProxy route's `Host` rewrite.

# Testing

Unit tests cover changed numeric ports, named ports, current-setting precedence for an unchanged backend, and Host rewrite recovery for previous backends.

/kind bug

**Release Note**

```release-note
Fixed endpoint probes for previous backends that use a different Service port or require host rewriting.
```

**Docs**

```docs
N/A
```


kahirokunn:
@linkvt Hi ✋ Could you please review this PR when you have time, if you don't mind? Thank you🙇

pkg/reconciler/contour/resources/kingress.go
```diff
@@ -31,6 +33,37 @@ import (
 	"knative.dev/pkg/logging"
 )
 
+type serviceIdentity struct {
+	name string
+	port string
+}
+
+func endpointProbeServices(ing *v1alpha1.Ingress) map[serviceIdentity]ServiceInfo {
```
linkvt, line 41: This func is quite similar to the previously used one: https://github.com/knative-extensions/net-contour/blob/577c757d150120d03d566e06d325a1e424d02df3/pkg/reconciler/contour/resources/httpproxy.go#L61-L81 

I'm wondering if we should maybe deduplicate this, we could pass a key function as parameter so that the map gets once keyed by the serviceIdentity and once by service name?

linkvt (commented):
Thanks for the PR, LGTM in general but I have no approver permissions here.

I left a comment to simplify things but its IMO optional

/lgtm

pkg/reconciler/contour/resources/kingress.go
kahirokunn: Thank you!
I've refactored this! 🙏

linkvt (commented):
/lgtm

Have no authority here though but still thanks :)

kahirokunn:
Thank you for your review! I really appreciate it!

kahirokunn:
@dprotaso @dsimansk Hi ✋ PTAL. Thank you 🙏

